PR #8, opened 2019-12-20
chore(test): rename DatastoreTest to ITDatastoreTest

BenWhitehead:
Integration tests are not ran on OS X CI builds, moving DatastoreTest &
LocalDatastoreHelperTest to work around the fact that the emulator is
accessible on kokoro nodes.


chingor13:
I believe datastore tests have an issue on our OSX test machine because it uses the emulator to run tests against. On OSX, IIRC, there's some kind of socket issue accepting connections. 

I think @kolea2 had investigated this for a bit.

BenWhitehead:
That's what it's looking like based on my debugging so far. But not having a mac I haven't been able to diagnose past that point yet.

kolea2:
Forgot I did this, I tried some WIP stuff but was unsuccessful: https://github.com/googleapis/google-cloud-java/pull/5070/files

PR #56, opened 2020-02-26
feat: expose reserveIds API

### LINKED ISSUE #53: [Datastore] expose 'reserveIds' API
Expose the  `'reserveIds'` API  to Prevents the supplied keys' IDs from being auto-allocated by Cloud Datastore.

- [ ] [reserveIds](https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/reserveIds)

/cc: @BenWhitehead

athakor:
Fixes #53 

Proposed Design [Doc](https://docs.google.com/document/d/1FYfg3-R6-AoWbv5JWlYoJYtG-L-JpQCpISy3IVDBpK0/edit#).

~/cc: @BenWhitehead

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
@@ -17,7 +17,6 @@
 package com.google.cloud.datastore;
 
 import com.google.cloud.Service;
-import com.google.datastore.v1.ReserveIdsResponse;
 import com.google.datastore.v1.TransactionOptions;
 import java.util.Iterator;
 import java.util.List;
@@ -170,7 +169,7 @@ public interface Datastore extends Service<DatastoreOptions>, DatastoreReaderWri
   List<Key> allocateId(IncompleteKey... keys);
 
   /**
-   * Reserve a list of IDs from a multiple complete key
+   * Reserve one or more keys, preventing them from being automatically allocated by Datastore.
    *
    * <p>Example of reserving multiple ids in a single batch.
    *
@@ -178,13 +177,13 @@ public interface Datastore extends Service<DatastoreOptions>, DatastoreReaderWri
    * KeyFactory keyFactory = datastore.newKeyFactory().setKind("MyKind");
    * Key key1 = keyFactory.newKey(10);
    * Key key2 = keyFactory.newKey("name");
-   * ReserveIdsResponse response = datastore.reserveIds(key1, key2);
+   * List<Key> keys = datastore.reserveIds(key1, key2);
    *
    * }</pre>
    *
    * @throws DatastoreException upon failure
    */
-  ReserveIdsResponse reserveIds(Key... keys);
+  List<Key> reserveIds(Key... keys);
 
   /**
    * {@inheritDoc}
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
@@ -170,7 +170,7 @@ public interface Datastore extends Service<DatastoreOptions>, DatastoreReaderWri
   List<Key> allocateId(IncompleteKey... keys);
 
   /**
-   * Reserve a list of IDs from a complete key
+   * Reserve a list of IDs from a multiple complete key
    *
    * <p>Example of reserving multiple ids in a single batch.
    *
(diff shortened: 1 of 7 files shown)


athakor:
@BenWhitehead PTAL on this and let me know your feedback.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
```diff
@@ -168,6 +169,23 @@
    */
   List<Key> allocateId(IncompleteKey... keys);
 
+  /**
+   * Reserve a list of IDs from a multiple complete key
```
BenWhitehead:
```suggestion
   * Reserve one or more keys, preventing them from being automatically allocated by Datastore
```

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -400,6 +402,36 @@ protected Entity computeNext() {
     }
   }
 
+  @Override
+  public ReserveIdsResponse reserveIds(Key... keys) {
+    ReserveIdsRequest.Builder requestPb = ReserveIdsRequest.newBuilder();
+    for (Key key : keys) {
+      if (key instanceof Key) {
+        key = (Key) IncompleteKey.fromPb(key.toPb());
```
BenWhitehead: Why do we need the `instanceof` check and Key -> pb -> IncompleteKey -> Key dance here?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
```diff
@@ -168,6 +169,23 @@
...
+   * KeyFactory keyFactory = datastore.newKeyFactory().setKind("MyKind");
+   * Key key1 = keyFactory.newKey(10);
+   * Key key2 = keyFactory.newKey("name");
+   * ReserveIdsResponse response = datastore.reserveIds(key1, key2);
+   *
+   * }</pre>
+   *
+   * @throws DatastoreException upon failure
+   */
+  ReserveIdsResponse reserveIds(Key... keys);
```
BenWhitehead: Can we change the return type of this method? I don't think we want to pass the underlying proto object out of this method.  Instead I think we can return the list of reserved ids as a `List<Key>`.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
athakor: done

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
athakor: yes,we can explicitly return `List<Key>` because `reserveIds` APi return the empty response body,if its successful.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
athakor: thanks for pointing out this, it's look like unused, I have already simplified this.

athakor:
@BenWhitehead all the comment has been addressed. PTAL

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -400,6 +401,41 @@ protected Entity computeNext() {
     }
   }
 
+  @Override
+  public List<Key> reserveIds(Key... keys) {
+    ReserveIdsRequest.Builder requestPb = ReserveIdsRequest.newBuilder();
+    for (Key key : keys) {
+      IncompleteKey incompleteKey = IncompleteKey.fromPb(key.toPb());
+      requestPb.addKeys(incompleteKey.toPb());
```
BenWhitehead: I don't think the conversion to an `IncompleteKey` here isn't necessary and can be simplified
```suggestion
      requestPb.addKeys(key.toPb());
```

Both `Key` and `IncompleteKey` use the same `toPb` on `BaseKey`.

BenWhitehead:
One small change, after that this will be ready to merge.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
BenWhitehead: Yeah, ReserveIdsResponse is empty and I thought it'd be more useful to users for us to take care of returning the ids that were reserved on success rather than them having to keep track of it out of band themselves. Thanks for updating.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
athakor: done

BenWhitehead:
Thanks for the feature @athakor!

PR #82, opened 2020-03-26
chore: Migrate datastore synth.py to bazel

vam-google:
The code was regenerated, there were no changes (except synth.metadata, which we agree to not include in these PRs)


vam-google:
@chingor13 PTAL

PR #108, opened 2020-04-20
feat: add flatten plugin to client and proto modules

stephaniewang526:


.gitignore
```diff
@@ -3,3 +3,4 @@ target/
 *.iml
 __pycache__/
 
+.flattened-pom
```
BenWhitehead, line 6: nit: `.flattened-pom.xml`

proto-google-cloud-datastore-v1/pom.xml
```diff
@@ -24,6 +24,16 @@
...
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>flatten-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
+
   <profiles>
     <profile>
```
BenWhitehead, line 38: Can you add the version tag to each of the java9 profiles in the repo to avoid the bug we ran into yesterday with java 9+?

.gitignore
stephaniewang526: Done

proto-google-cloud-datastore-v1/pom.xml
stephaniewang526: Done.

BenWhitehead:
Thanks @stephaniewang526!

PR #213, opened 2020-09-01
chore(build): update synth config to generate README.md

BenWhitehead:


BenWhitehead:
Good catch, I've pulled out the change to `synth.metadata`.

PR #506, opened 2021-08-20
feat: Change checksum algorithm to CRC32C. And localize checksum handling t…

vnorigoog:
…o one file.

### DIFF
--- a/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/EndToEndChecksumHandler.java
+++ b/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/EndToEndChecksumHandler.java
@@ -16,8 +16,9 @@
 package com.google.datastore.v1.client;
 
 import com.google.api.client.http.HttpResponse;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
+import com.google.common.hash.HashCode;
+import com.google.common.hash.Hasher;
+import com.google.common.hash.Hashing;
 
 /** This class provides End-to-End Checksum API for http protocol. */
 class EndToEndChecksumHandler {
@@ -25,22 +26,25 @@ class EndToEndChecksumHandler {
   static final String HTTP_REQUEST_CHECKSUM_HEADER = "x-request-checksum-348659783";
   /** The checksum http header on http responses */
   static final String HTTP_RESPONSE_CHECKSUM_HEADER = "x-response-checksum-348659783";
-  /** Algorithm used for checksum */
-  private static final String MD5 = "MD5";
+
+  final Hasher hasher = EndToEndChecksumHandler.getNewCrc32cHasher();
 
   /**
    * Create and return checksum as a string value for the input 'bytes'.
    *
    * @param bytes raw message for which the checksum is being computed
    * @return computed checksum as a hex string
-   * @throws RuntimeException if MD5 Algorithm is not found in the VM
    */
   static String computeChecksum(byte[] bytes) {
     if (bytes == null || (bytes.length == 0)) {
       return null;
     }
-    return com.google.common.io.BaseEncoding.base16()
-        .encode(getMessageDigestInstance().digest(bytes));
+    HashCode hc = getNewCrc32cHasher().putBytes(bytes).hash();
+    return hc.toString();
+  }
+
+  private static Hasher getNewCrc32cHasher() {
+    return Hashing.crc32c().newHasher();
   }
 
   /**
@@ -58,14 +62,6 @@ class EndToEndChecksumHandler {
         && checksum.equalsIgnoreCase(computeChecksum(bytes));
   }
 
-  static MessageDigest getMessageDigestInstance() {
-    try {
-      return MessageDigest.getInstance(MD5);
-    } catch (NoSuchAlgorithmException e) {
-      throw new RuntimeException("MD5 algorithm is not found when computing checksum!");
-    }
-  }
-
   static boolean hasChecksumHeader(HttpResponse response) {
     String checksum = getChecksumHeader(response);
     return checksum != null && !checksum.isEmpty();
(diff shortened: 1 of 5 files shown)


datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/EndToEndChecksumHandler.java
```diff
@@ -16,31 +16,35 @@
...
     if (bytes == null || (bytes.length == 0)) {
       return null;
     }
-    return com.google.common.io.BaseEncoding.base16()
-        .encode(getMessageDigestInstance().digest(bytes));
+    HashCode hc = getNewHasher().putBytes(bytes).hash();
+    return hc.toString();
+  }
+
+  private static Hasher getNewHasher() {
```
kolea2, line 46: for better readability, I would rename this to `getNewCrc32cHasher`

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/EndToEndChecksumHandler.java
```diff
@@ -74,4 +70,12 @@ static boolean hasChecksumHeader(HttpResponse response) {
   static String getChecksumHeader(HttpResponse response) {
     return response.getHeaders().getFirstHeaderStringValue(HTTP_RESPONSE_CHECKSUM_HEADER);
   }
+
+  void update(byte[] bytes, int off, int len) {
```
kolea2, line 74: should these be `protected`?

kolea2 (changes requested):

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/EndToEndChecksumHandler.java
vnorigoog: Done

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/EndToEndChecksumHandler.java
vnorigoog: These 2 methods - update(..), hash() - are used by ChecksumEnforcingInputStream.
thats why I had to make them package scoped.

vnorigoog (commented):
Thanks for the review.. I missed looking at the review comments for the last 1 week! sorry about the delay

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/RemoteRpcTest.java
```diff
@@ -174,7 +174,7 @@ public void testHttpHeaders_expectE2eChecksumHeader() throws IOException {
         httpRequest
             .getHeaders()
             .getFirstHeaderStringValue(EndToEndChecksumHandler.HTTP_REQUEST_CHECKSUM_HEADER);
-    assertEquals(32, header.length());
```
kolea2, line 177: sorry I missed this in the previous review, can you confirm why this is changing and if it's expected?

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/RemoteRpcTest.java
```diff
@@ -174,7 +174,7 @@ public void testHttpHeaders_expectE2eChecksumHeader() throws IOException {
...
+    assertEquals(8, header.length());
```
crwilcox, line 177: This seems an odd change. Is the crc32c hash that much smaller?

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/ChecksumEnforcingInputStreamTest.java
```diff
@@ -66,9 +63,7 @@ public void read_withInvalidChecksum() {
     // Since checksum should be correct, do not expect IOException
     try (ChecksumEnforcingInputStream instance =
         new ChecksumEnforcingInputStream(
-            new ByteArrayInputStream("hello there".getBytes(UTF_8)),
-            "this checksum is invalid",
-            digest)) {
```
crwilcox, line 71: it seems odd to me that there is no digest. Is this different than the checksum from crc32c? Is this just internal now somehow?

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/RemoteRpcTest.java
vnorigoog: MD5 is 32 bytes long
since we are changing to CRC32c - which is 8 bytes long - this change reflects that..

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/RemoteRpcTest.java
vnorigoog: CRC32C is 8 bytes only.

I would have preferred to use MD5 but internal review with ESF team pointed out that CRC32C is preferred (partly because ESF already uses CRC32C for internal checksum computation and this change would help cut down on rec-computation of checksum)

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/ChecksumEnforcingInputStreamTest.java
vnorigoog: yes sir - CRC32C doesn't have/need digest..

Chris, Kristen - you know you guys are *most* welcome to recommend MD5, instead of CRC32C, for clients.
MD5 is way stronger and is readily available in all Java versions (whereas CRC32C is not in core java libs until JDK 11)

vnorigoog (commented):
Thanks for the reviews

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/RemoteRpcTest.java
crwilcox: we have also recently been moving to crc32c for other libs, as the computation should be faster for streams.

vnorigoog: oh ok - then moving to CRC32C in this client is probably acceptable.

vnorigoog:
appreciate follow-up feedback on the PR please

kolea2 (approved):
lgtm, will let @crwilcox take another look

PR #584, opened 2021-12-02
chore: migrate to owlbot

Neenu1995:


eaball35 (approved):

Neenu1995:
Merge after Kristen reviews.

versions.txt
```diff
@@ -3,7 +3,5 @@
 
 google-cloud-datastore:2.2.0:2.2.1-SNAPSHOT
 google-cloud-datastore-bom:2.2.0:2.2.1-SNAPSHOT
-google-cloud-datastore-parent:2.2.0:2.2.1-SNAPSHOT
```
kolea2, line 6: @Neenu1995 can you take a look at why this was deleted?

kolea2 (changes requested):

versions.txt
Neenu1995: the parent version is same as the google-cloud-library version. Doesn't have to be tracked on versions.txt

kolea2 (approved):

PR #591, opened 2021-12-05
style: Fix 2 ErrorProneStyle findings

### LINKED ISSUE #590: Monorepo style suggested changes
Internally to Google, clshepherd / clrobot suggest changes to apply best practices. It made the following suggestions:

> Fix 2 ErrorProneStyle findings:
> * Constructors and methods with the same name should appear sequentially with no other code in between. Please re-order or re-name methods.
> * This catch block catches an exception and re-throws another, but swallows the caught exception rather than setting it as a cause. This can make debugging harder.

I'm not sure these are all available via the public errorprone module, but perhaps some would get caught upstream if it were installed per https://errorprone.info/docs/installation#maven.

Capstan:
* Constructors and methods with the same name should appear sequentially with no other code in between.
* This catch block catches an exception and re-throws another, but swallows the caught exception rather than setting it as a cause. This can make debugging harder.

Courtesy of clshepherd/clrobot in the monorepo.

Fixes #590.

### DIFF
--- a/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/DatastoreHelper.java
+++ b/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/DatastoreHelper.java
@@ -483,6 +483,16 @@ public final class DatastoreHelper {
     return Value.newBuilder().setTimestampValue(toTimestamp(date.getTime() * 1000L));
   }
 
+  /** Makes a GeoPoint value. */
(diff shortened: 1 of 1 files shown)


suztomo (dismissed):

kolea2 (approved):

alicejli:
going to open and reclose in attempt to retry the stuck tests.

PR #592, opened 2021-12-05
style: Add errorprone warnings.

Capstan:
This introduces use of the errorprone plugin as described at https://errorprone.info/docs/installation#maven. The warnings can then be addressed, `@SuppressWarnings(…)`ed, or if desired, have entire classes of warnings turned off. Then once the codebase is clean, the warnings can be treated as errors. This will reduce the drift between github and the monorepo, which generally tries to enforce errorprone cleanliness.

Related to #590.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,7 @@
     <github.global.server>github</github.global.server>
     <site.installationModule>google-cloud-datastore-parent</site.installationModule>
     <project.javadoc.protobufBaseURL>https://googleapis.dev/java/google-api-grpc/latest</project.javadoc.protobufBaseURL>
+    <error-prone.version>2.10.0</error-prone.version>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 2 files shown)


Capstan:
I've adapted my original PR to use profiles to only apply ErrorProne to JDK 9 and beyond. I don't think the GraalVM test failure is my doing, as it appears to be missing some configuration file I've not touched. I believe this is ready for review.

google-cloud-datastore/pom.xml
```diff
@@ -16,6 +16,7 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-datastore</site.installationModule>
+    <error-prone.version>2.10.0</error-prone.version>
```
kolea2, line 19: is there any particular reason why this is declared in these two modules? It would be cleaner to declare this in the parent and have each module inherit and run the config: https://github.com/googleapis/java-datastore/blob/main/pom.xml

google-cloud-datastore/pom.xml
```diff
@@ -110,4 +111,54 @@
...
+                <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
+                <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
+              </compilerArgs>
+              <annotationProcessorPaths>
+                <path>
+                  <groupId>com.google.errorprone</groupId>
+                  <artifactId>error_prone_core</artifactId>
+                  <version>${error-prone.version}</version>
+                </path>
+                <!-- Other annotation processors go here.
```
kolea2, line 152: is this comment relevant? Can we clean up?

google-cloud-datastore/pom.xml
kolea2: to be clear, I am referring to both the property definition as well as plugin definition.

google-cloud-datastore/pom.xml
Capstan: I can remove it; I didn't know if you'd all have a need for annotation processors so it was at least a nudge to check.

google-cloud-datastore/pom.xml
Capstan: Honestly, I was afraid it would affect the ability to run `mvn` in the child directory and I didn't know whether you wanted to preserve that behavior. I can centralize these settings.

kolea2: Typically all CI is run from the parent, so let's try centralizing it for consistency and see if we run into any issues.

google-cloud-datastore/pom.xml
kolea2: makes sense! I'd prefer to remove just to reduce some noise.

google-cloud-datastore/pom.xml
Capstan: Done.

google-cloud-datastore/pom.xml
Capstan: Done!

Capstan:
*ping*

suztomo (approved):

Capstan:
Oh, so close 😮

pom.xml
```diff
@@ -210,6 +211,51 @@
...
+        <jdk>[9,)</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <version>3.8.0</version>
+            <configuration>
+              <source>8</source>
```
kolea2, line 227: just curious if we need this since this profile is for jdk 9+ only?

kolea2 (approved):
LGTM with one remaining question

pom.xml
Capstan: Ah, possibly not. I think I can just use

```
<configuration>
  <release>9</release>
</configuration>
```

Lemme try that.

Capstan: Latest version uses that and seems to be fine with it.

kolea2: I think you can simply remove it? I was looking at https://github.com/googleapis/java-pubsublite/blob/adb22207fb3cba10de95a5e6399456d362408dd7/pom.xml#L166-L199

Capstan: I was under the understanding if you didn't have either `<source>` or `<target>` or `<release>`, the default behavior was that it was treated as Java 6 (i.e., 1.6), but perhaps that was in earlier releases of Java? Anyway, removing it also works.

kolea2 (approved):
LGTM, thanks!

PR #611, opened 2022-01-08
feat: add DatastoreAdmin client

sorced-jim:


README.md
```diff
@@ -213,15 +213,15 @@ Testing
 
 This library has tools to help write tests for code that uses the Datastore.
 
-See [TESTING.md](https://github.com/googleapis/google-cloud-java/blob/main/TESTING.md#testing-code-that-uses-datastore) to read more about testing.
+See [TESTING.md](https://github.com/googleapis/google-cloud-java/blob/master/TESTING.md#testing-code-that-uses-datastore) to read more about testing.
```
kolea2, line 216: I think these are reverting updates to the readme - can you remove these changes?

synth.metadata
```diff
@@ -4,26 +4,35 @@
       "git": {
         "name": ".",
         "remote": "https://github.com/googleapis/java-datastore.git",
-        "sha": "e08f9d93d8a8b912fffd602bed3189c1bed664a6"
+        "sha": "8985e685d8ce3d8c00d6d34ff13333bac6516b5f"
```
kolea2: I don't think this should be deleted

kolea2 (changes requested):
Can we revert any non admin changes? I see a lot of java proto-google-cloud-datastore-v1 updates that are adding a bunch of noise

synth.py
```diff
@@ -40,31 +40,83 @@
...
 
-for version in versions:
+def generate_client(service, version, proto_path=None, bazel_target=None, package=None):
   library = gapic.java_library(
       service=service,
       version=version,
-      bazel_target=f'//google/{service}/{version}:google-cloud-{service}-{version}-java',
-      proto_path=f'google/{service}/{version}',
+      proto_path=proto_path if proto_path else f'google/{service}/{version}',
+      bazel_target=bazel_target if bazel_target else f'//google/{service}/{version}:google-cloud-{service}-{version}-java',
```
kolea2: can we swap the order back to what it was to reduce the diff?

synth.py
```diff
@@ -40,31 +40,83 @@
...
 
   s.replace(
       library / f'proto-google-cloud-{service}-{version}-java/src/**/*.java',
       protobuf_header,
       f'{license_header}{protobuf_header}'
   )
 
-  # Datastore only generates protos
-  s.copy(library / f'proto-google-cloud-{service}-{version}-java/src', f'proto-google-cloud-{service}-{version}/src')
+  if service == "datastore-admin":
```
kolea2: These fixes can be refactored, see https://github.com/googleapis/java-bigtable/blob/main/synth.py#L123-L125 (fix_proto_headers and fix_grpc_headers).

kolea2 (changes requested):

PR #686, opened 2022-03-17
chore: Change the Codeowner to cloud-native-db-dpes

Mariatta:
Replaced `@googleapis/firestore-dpe` with `@googleapis/cloud-native-db-dpes`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,9 +4,9 @@
 # For syntax help see:
 # https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-# The @googleapis/cloud-native-db-dpes is the default owner for changes in this repo
-*                       @googleapis/yoshi-java @googleapis/cloud-native-db-dpes
-**/*.java               @googleapis/cloud-native-db-dpes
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -4,9 +4,9 @@
 # For syntax help see:
 # https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-# The @googleapis/firestore-dpe is the default owner for changes in this repo
-*                       @googleapis/yoshi-java @googleapis/firestore-dpe
-**/*.java               @googleapis/firestore-dpe
+# The @googleapis/cloud-native-db-dpes is the default owner for changes in this repo
+*                       @googleapis/yoshi-java @googleapis/cloud-native-db-dpes
```
kolea2, line 8: we should have both teams rather than replace them

kolea2 (changes requested):

.github/CODEOWNERS
Mariatta: The `googleapis/firestore-dpe` team doesn't seem to exist 😅  that's why we're replacing it with the `cloud-native-db-dpes` team.

kolea2: oops sorry looked too quickly! That should be replaced with `api-firestore`

kolea2 (approved):

PR #712, opened 2022-04-18
feat: add ReadOption.ReadTime to support timestamp reads.

yixiaoshen:

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOption.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOption.java
@@ -49,8 +49,8 @@ public abstract class ReadOption implements Serializable {
   }
 
   /**
-   * Reads entities as they were at the given time. This may not be older than 270 seconds.
-   * This value is only supported for Cloud Firestore in Datastore mode.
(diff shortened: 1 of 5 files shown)


kolea2 (commented):
looks good, can we also add an integration test?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOption.java
```diff
@@ -57,6 +77,14 @@ public static EventualConsistency eventualConsistency() {
     return new EventualConsistency(true);
   }
 
+  /**
+   * Returns a {@code ReadOption} that specifies read time, allowing Datastore to return results
+   * from lookups and queries at a particular timestamp.
+   */
+  public static ReadTime readTime(Timestamp time) {
```
kolea2, line 84: please add `@BetaApi`

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOption.java
yixiaoshen: done

kolea2 (approved):

PR #723, opened 2022-04-21
fix(java): rename test to be picked up by native profile

mpeddada1:
A couple of changes to note:
1) The maven-surefire-plugin defined in [datastore-v1-proto-client/pom.xml](https://github.com/googleapis/java-datastore/pull/723/files#diff-18fb0d587c1b12f0cb5729d02ed0d791ec5b5ea14844da40e21dbf9cbc681908)  was overwriting the one defined in the native profile brought in from java-shared-config, therefore, resulting in the following error: ```Test configuration file wasn't found. Make sure that test execution wasn't skipped.```.  Tested this change locally to verify that integration tests are excluded when unit tests are run. Also, java-shared-config brings in the same config defined in the POM: https://github.com/googleapis/java-shared-config/blob/b5e05c6726c60a7ad7c0d9588b4fab69a0b4e6b7/pom.xml#L87

2) Replaced usages of `@Rule public ExpectedException thrown = ExpectedException.none();` with Truth. (background: go/cloud-java-native-image#truth-vs-hamcrest-use-truth)

3) Renamed test that uses EasyMock `ITDatastoreTest` -> `DatastoreTest` as mocking is incompatible with native image compilation atm.

### DIFF
--- a/datastore-v1-proto-client/src/main/resources/META-INF/native-image/native-image.properties
+++ b/datastore-v1-proto-client/src/main/resources/META-INF/native-image/native-image.properties
@@ -1,0 +1,1 @@
+Args = --enable-url-protocols=https,http
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## [2.4.0](https://github.com/googleapis/java-datastore/compare/v2.3.1...v2.4.0) (2022-04-21)
+
+
+### Features
+
+* expose new read_time API fields, currently only available in private preview ([#706](https://github.com/googleapis/java-datastore/issues/706)) ([e11bd20](https://github.com/googleapis/java-datastore/commit/e11bd20fb99a3d671b09e1b46247fdfd11927fbe))
+* next release from main branch is 2.4.0 ([#715](https://github.com/googleapis/java-datastore/issues/715)) ([9cc80c9](https://github.com/googleapis/java-datastore/commit/9cc80c999d08d5ccfd667e765f94181904e17645))
+
+
+### Dependencies
+
+* update dependency com.google.cloud:google-cloud-shared-dependencies to v2.10.0 ([#710](https://github.com/googleapis/java-datastore/issues/710)) ([ed8de9c](https://github.com/googleapis/java-datastore/commit/ed8de9cc519d5817b8af2c9705b9f335b909b78a))
+* update dependency com.google.cloud:native-image-support to v0.13.1 ([#709](https://github.com/googleapis/java-datastore/issues/709)) ([354ad90](https://github.com/googleapis/java-datastore/commit/354ad90aa075b203329b9ebde88f2b668351c37e))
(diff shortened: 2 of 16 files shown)


mpeddada1:
The Kokoro Native Image job is now passing

mpeddada1:
However, the unit tests seem to be failing with the following error:
```
Trying to download...
Downloading Java 17.0.3+7 (Zulu) from https://cdn.azul.com/zulu/bin/zulu17.34.19_1-ca-jdk17.0.3-linux_x64.tar.gz ...
Error: Unexpected HTTP response: 404
```

datastore-v1-proto-client/src/main/resources/META-INF/native-image/native-image.properties
```diff
@@ -0,0 +1 @@
+Args = --enable-url-protocols=https,http
```
suztomo, line 1: Does this library use HTTP (not HTTPS) somewhere in users' environment? (What's the error without this argument?)

If the request is only for testing, then this file should be under test.

suztomo:
@kolea2 This is a followup from our conversation after the Wednesday meeting.

datastore-v1-proto-client/src/main/resources/META-INF/native-image/native-image.properties
mpeddada1: Looking at the stacktrace, this is coming from `com.google.datastore.v1.client.RemoteRpc`: 
```
 Caused by: java.lang.IllegalArgumentException: java.net.MalformedURLException: Accessing an URL protocol that was not enabled. The URL protocol https is supported but not enabled by default. It must be enabled by adding the --enable-url-protocols=https option to the native-image command.
       com.google.api.client.http.GenericUrl.parseURL(GenericUrl.java:679)
       com.google.api.client.http.GenericUrl.<init>(GenericUrl.java:125)
       com.google.api.client.http.GenericUrl.<init>(GenericUrl.java:108)
       com.google.datastore.v1.client.RemoteRpc.resolveURL(RemoteRpc.java:161)
       com.google.datastore.v1.client.RemoteRpc.<init>(RemoteRpc.java:62)
```
Normally this configuration would come from gax, but this module doesn't import it so we had to add this setting in.

suztomo (approved):
If the allow HTTP is passed in GAX already, then no need to reject it here.

datastore-v1-proto-client/src/main/resources/META-INF/native-image/native-image.properties
mpeddada1: The error is coming from `RemoteRpc.resolveURL` but it's used in [`RemoteRpc#call` ]( https://github.com/googleapis/java-datastore/blob/957a6b5b940edc476344a38cc6038a4a4851039f/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java#L86) which is later referenced in `Datastore`. For this reason, I think we would still need it in the main configuration?

kolea2 (approved):

mpeddada1:
Thank you all so much!

PR #748, opened 2022-05-25
feat: Expose Datastore Admin generated client

pcostell:


grpc-google-cloud-datastore-v1/src/main/java/com/google/datastore/v1/DatastoreGrpc.java
```diff
@@ -0,0 +1,1053 @@
+/*
```
kolea2, line 1: I am surprised to see this get generated for only the admin client - is this correct?

.repo-metadata.json
```diff
@@ -13,7 +13,5 @@
   "api_id": "datastore.googleapis.com",
   "library_type": "GAPIC_COMBO",
   "api_description": "is a fully managed, schemaless database for\nstoring non-relational data. Cloud Datastore automatically scales with\nyour users and supports ACID transactions, high availability of reads and\nwrites, strong consistency for reads and ancestor queries, and eventual\nconsistency for all other queries.",
-  "excluded_dependencies": "grpc-google-cloud-datastore-v1",
```
kolea2, line 16: same-ish question here, why are we no longer excluding this for admin-only?

grpc-google-cloud-datastore-v1/src/main/java/com/google/datastore/v1/DatastoreGrpc.java
pcostell: Owl seemed to be trying to add the repo as a dependency, but I don't see any reason that is actually true. I may have just misconfigured the owl configuration during that run. Removed it, we'll see if owl tries to put it back.

pcostell:
I think actually ready for review now @kolea2, thanks for your help!

pom.xml
```diff
@@ -156,6 +156,26 @@
...
+        <version>2.5.2-SNAPSHOT</version><!-- {x-version-update:grpc-google-cloud-datastore-v1:current} -->
+      </dependency>
+      <dependency>
+        <groupId>com.google.api.grpc</groupId>
+        <artifactId>grpc-google-cloud-datastore-admin-v1</artifactId>
+        <version>2.5.2-SNAPSHOT</version><!-- {x-version-update:grpc-google-cloud-datastore-admin-v1:current} -->
+      </dependency>
+      <dependency>
+        <groupId>com.google.api.grpc</groupId>
+        <artifactId>grpc-google-location</artifactId>
```
kolea2, line 176: what's this? I think the version should also be something different?

kolea2 (commented):
looks great, one comment/question. Thanks!

pom.xml
pcostell: Definitely should not be a required dependency. Removed it.

kolea2 (approved):

fkorotkov:
Can we do the same for the regular Datastore client? There is no `grpc-google-cloud-datastore-v1` unfortunetly.

PR #790, opened 2022-07-06
docs(sample): Add a top-level Datastore samples README

### LINKED ISSUE #745: What are the steps to run samples/snippets/src/main/java/com/google/datastore/snippets/TaskList.java ?
Thanks for stopping by to let us know something could be better!

**Is your feature request related to a problem? Please describe.**
What the problem is. Example: I'm unable to run TaskList.java.

**Describe the solution you'd like**
We previously documented the steps to run the sample [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/8a6966fbf71a9629997cd381f397460032768e78/datastore/README.md)

I'd like to see a similar README restored. Someone suggested that a pom.xml file might also be missing.

bw-yuhogg:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-datastore/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea. **Issue already existed**
- [x] Ensure the tests and linter pass **N/A**
- [x] Code coverage does not decrease (if any source code was changed) **N/A**
- [x] Appropriate docs were updated (if necessary) **Strictly a docs addition**

Fixes #745 ☕️

Note: I can confirm that the existing pom.xml works correctly.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -1,0 +1,30 @@
+# Getting Started with Datastore Samples
+
+## Running a sample using Cloud Shell
+
+The Google Cloud Shell has application default credentials from its compute instance which will allow you to run samples without having to obtain `GOOGLE_APPLICATION_CREDENTIALS`.
+Go to the [Datastore Client Readme](https://github.com/googleapis/java-datastore#samples) to run each sample in the Cloud Shell.
+
+## Running a sample at the command line
+
+1. First set up the `GOOGLE_CLOUD_PROJECT` environment variable.
+
+2. Obtain authentication credentials.
+
+    Create local credentials by running the following command and following the
+    oauth2 flow (read more about the command [here][auth_command]):
+
+        gcloud auth application-default login
+
+    Or manually set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
+    to point to a service account key JSON file path.
+
+    Learn more at [Setting Up Authentication for Server to Server Production Applications][ADC].
+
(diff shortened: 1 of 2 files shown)


samples/README.md
```diff
@@ -0,0 +1,14 @@
+# Getting Started with Datastore Samples
+
+## Running a sample using Cloud Shell
+
+The Google Cloud Shell has application default credentials from its compute instance which will allow you to run samples without having to obtain `GOOGLE_APPLICATION_CREDENTIALS`.
+Go to the [Datastore Client Readme](https://github.com/googleapis/java-datastore#samples) to run each sample in the Cloud Shell.
+
+## Running a sample at the command line
+
+First set up `GOOGLE_APPLICATION_CREDENTIALS` and `GOOGLE_CLOUD_PROJECT` environment variables before running any samples.
```
lesv, line 10: The shouldn't need to set `GOOGLE_APPLICATION_CREDENTIALS` if they use `gcloud auth application-default login`

lesv (approved):
Consider the comment

samples/README.md
rwhogg: Fixed, thanks!

kolea2 (approved):

kolea2 (changes requested):
left one last comment on thread

kolea2 (approved):

PR #823, opened 2022-09-14
feat: Count API 

jainsahab:
Add support for querying the Datastore backend to just get the count (i.e. the size of the result set) for a query, instead of getting the entire result set.


google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResults.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Objects;
+
+public class AggregationResults implements Iterable<AggregationResult> {
+
+  private final List<AggregationResult> aggregationResults;
+  private Timestamp readTime;
+
+  public AggregationResults(List<AggregationResult> aggregationResults,
```
gforgurups, line 29: As its public constructor we may have to check for null condition for aggregationResults else all method may fail with unexpected behavior.

gforgurups (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQueryProtoPreparer.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.cloud.datastore.StructuredQuery.OrderBy;
+import com.google.cloud.datastore.execution.request.ProtoPreparer;
+import com.google.datastore.v1.Query;
+import com.google.protobuf.Int32Value;
+
+public class StructuredQueryProtoPreparer implements ProtoPreparer<StructuredQuery<?>, Query> {
+
+  @Override
+  public Query prepare(StructuredQuery<?> query) {
+    com.google.datastore.v1.Query.Builder queryPb = com.google.datastore.v1.Query.newBuilder();
```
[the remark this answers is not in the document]
jainsahab: Do you mean `query` (method parameter) ?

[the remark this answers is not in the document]
gforgurups: Yes that's right. In addition, is com.google.datastore.v1.Query.newBuilder() is considered safe enough not to return null?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
```diff
@@ -0,0 +1,117 @@
...
+  public static class Builder {
+
+    private String namespace;
+    private Mode mode;
+    private final List<Aggregation> aggregations;
+    private StructuredQuery<?> nestedStructuredQuery;
+    private GqlQuery<?> nestedGqlQuery;
+
+    public Builder() {
+      this.aggregations = new ArrayList<>();
```
gforgurups, line 76: Just a food for thought. As aggregations are applied on raw data and we are not handling nested aggregations , I presume we don't want to allow duplicate aggregations. If previous aggregation results are not cached than duplicating same aggregation operation on huge record set can be costly. So, should we consider using Set Interface (Hashset)?

.gitignore
```diff
@@ -3,4 +3,5 @@ target/
 *.iml
 __pycache__/
 
-.flattened-pom.xml
\ No newline at end of file
+.flattened-pom.xml
+*.patch
```
kolea2, line 7: was this change intentional?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
```diff
@@ -494,26 +511,17 @@ public boolean equals(Object obj) {
...
 
   @Override
-  void populatePb(com.google.datastore.v1.RunQueryRequest.Builder requestPb) {
+  public void populatePb(com.google.datastore.v1.RunQueryRequest.Builder requestPb) {
     requestPb.setGqlQuery(toPb());
   }
 
   @Override
-  Query<V> nextQuery(com.google.datastore.v1.RunQueryResponse responsePb) {
+  public RecordQuery<V> nextQuery(com.google.datastore.v1.RunQueryResponse responsePb) {
```
kolea2, line 524: same here

google-cloud-datastore/src/main/java/com/google/cloud/datastore/RecordQuery.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.datastore;
+
+import com.google.api.core.InternalApi;
+import com.google.cloud.datastore.Query.ResultType;
+
+@InternalApi
+public interface RecordQuery<V>{
```
kolea2, line 22: is this still needed?

google-cloud-datastore/clirr-ignored-differences.xml
```diff
@@ -11,4 +11,9 @@
         <method>com.google.datastore.v1.ReserveIdsResponse reserveIds(com.google.datastore.v1.ReserveIdsRequest)</method>
         <differenceType>7012</differenceType>
     </difference>
+    <difference>
```
kolea2, line 14: as we are now on Java 8, please opt for a default method on the interface and remove this exception

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AggregationBuilder.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.datastore.aggregation;
+
+public interface AggregationBuilder<A extends Aggregation> {
```
kolea2, line 19: most builders in the library are simply static classes - do we need this level of abstraction?

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AggregationBuilder.java
jainsahab: Hi @kolea2 , It will be required while adding aggregation to a query.

`
Query<AggregationResult> aggregationQuery = Query.newAggregationQueryBuilder()
  .addAggregation(count().as("total"))
  .addAggregation(sum("cost").as("total_cost"))
  .over(query)
  .build();
`

In this case, both sum and count are returning the instances of AggregationBuilder, and aggregationQuery can use the builders without being aware of the concrete builder implementation.

If we don't do this, then we need to have `N `overloaded version of `addAggregation` method for `N` type of aggregation query we will be supporting in future.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public GqlQuery<?> getNestedGqlQuery() {
+    return nestedGqlQuery;
+  }
+
+  public Mode getMode() {
+    return mode;
+  }
+
+  public static class Builder {
```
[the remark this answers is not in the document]
jainsahab: Hi @gforgurups, In a way we can do that. But we need AggregationQuery's `Builder` due to following reasons:

- End user is accustomed to using builder (ex: Query.newEntityQueryBuilder, Query.newKeyQueryBuilder etc.), and we created a new `Query.newAggregationQueryBuilder` to access this builder class. If we expose AggregationQuery class directly we will be breaking the pattern established in their mind.
- If we see from code perspective, we have followed the existing convention and maintain the consistency with other Query classes by exposing this builder.

These two reasons are good enough for us to keep having this Builder, let me know if you think otherwise.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/CountAggregation.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.protobuf.Int64Value;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+public class CountAggregation extends Aggregation {
+
+  private final long limit;
+
+  private CountAggregation(String alias, long limit) {
```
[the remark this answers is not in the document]
jainsahab: that make sense @gforgurups . There is nothing wrong in giving creational access to Aggregations. I'll make this change and add one more version `addAggregation(Aggregation)`method along with `addAggregation(AggregationBuilder)`. so that user can pass Aggregation directly without it's builder.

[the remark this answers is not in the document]
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/5097cd69fc6c1684103fa4500c289b86fe81d1ce.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResults.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.api.core.InternalApi;
+import com.google.cloud.Timestamp;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Objects;
+
+public class AggregationResults implements Iterable<AggregationResult> {
+
+  private final List<AggregationResult> aggregationResults;
+  private Timestamp readTime;
```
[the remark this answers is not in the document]
jainsahab: Yes, fixed in the latest version of code.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQueryProtoPreparer.java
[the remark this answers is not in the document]
jainsahab: Hi @gforgurups 

For `query` parameter, we have already done the [validation here](https://github.com/googleapis/java-datastore/pull/823/files#diff-4d491d7fe61ca45f3587ec96ad5c619b7ec0c147f33b150b47d681195f814536R108-R109) at the creation time itself. At this point in time when we. are populating protos we are certain that it wont' be null.


and regarding `com.google.datastore.v1.Query.newBuilder()` is a generated code from our Proto definitions. Hence it is considered safe enough to not to return null.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/6508ff6dc97171ee652a8c9b189f7425e9d4bd45.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResults.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/6508ff6dc97171ee652a8c9b189f7425e9d4bd45.

.gitignore
jainsahab: I was maintaining multiple patches in my local, so thought it would be useful. Do you want me to remove this ?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/RecordQuery.java
jainsahab: Hi @kolea2 , before this feature, `Query` (top level class) was used heavily in the code to generalise between GqlQuery and StructuredQuery,  We used `Query` reference so that different components can still remain oblivious to the underlying Query. 

With the arrival of AggregationQuery, we had to move entity records specific behaviour to `RecordQuery` to keep the `Query` clean enough to be the top level class. Instead of using `Query` reference, existing query handling mechanism uses `RecordQuery` so that existing code doesn't change much.

If we dont do this, then we need to write code to handle StructuredQuery and GqlQuery separately.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
```diff
@@ -494,26 +511,17 @@ public boolean equals(Object obj) {
...
-      queryPb.addPositionalBindings(argument.toPb());
-    }
-    return queryPb.build();
+    GqlQueryProtoPreparer protoPreparer = new GqlQueryProtoPreparer();
+    return protoPreparer.prepare(this);
   }
 
   @Override
-  void populatePb(com.google.datastore.v1.RunQueryRequest.Builder requestPb) {
+  public void populatePb(com.google.datastore.v1.RunQueryRequest.Builder requestPb) {
```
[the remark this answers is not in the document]
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/5f1e419a3e30aca687e49eadccd0a135a8f1b2e0.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/5f1e419a3e30aca687e49eadccd0a135a8f1b2e0.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
```diff
@@ -0,0 +1,117 @@
...
+import static com.google.cloud.datastore.AggregationQuery.Mode.STRUCTURED;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.cloud.datastore.aggregation.Aggregation;
+import com.google.cloud.datastore.aggregation.AggregationBuilder;
+import java.util.ArrayList;
+import java.util.List;
+
+public class AggregationQuery extends Query<AggregationResults> {
```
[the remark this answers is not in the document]
jainsahab: ✅  Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/5e2fcb3d4aeee19181d50d6b64b4d54d7363a95b.

google-cloud-datastore/clirr-ignored-differences.xml
kolea2: I am still in favor of adding a default method rather than adding a breaking change here - these other exceptions (I believe) were when we were still on Java 7.

jainsahab: Just being the devil's advocate here and contradicting the `clirr` check. How is adding a  new method can be considered a breaking change? We are not breaking any existing expectations of an end user here, Just providing them with more capabilities than the older versions.

jainsahab: BTW, i quickly added a dummy default method in the above mentioned interface, and clirr still failed.
image

google-cloud-datastore/src/test/java/com/google/cloud/datastore/AggregationQueryTest.java
```diff
@@ -0,0 +1,155 @@
...
+      .build();
+
+  @Rule
+  public ExpectedException exceptionRule = ExpectedException.none();
+
+  @Test
+  public void testAggregations() {
+    AggregationQuery aggregationQuery = Query.newAggregationQueryBuilder()
+        .setNamespace(NAMESPACE)
+        .addAggregation(new CountAggregation("total_upto_100", 100))
```
vishwarajanand, line 52: The [design doc](http://go/java-datastore-count-design#heading=h.nxddek4w5dli) mentions this:

```
  .addAggregation(count().as("total"))           // not CountAggregation
```

jainsahab: Yes, both syntax are supported.

https://github.com/googleapis/java-datastore/pull/823#discussion_r971588252

The next test in the same file follows the builder construct.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/AggregationResultTest.java
```diff
@@ -0,0 +1,36 @@
...
+public class AggregationResultTest {
+
+  @Test
+  public void shouldGetAggregationResultValueByAlias() {
+    AggregationResult aggregationResult = new AggregationResult(ImmutableMap.of(
+        "count", LongValue.of(45),
+        "count_upto_30", LongValue.of(30)
+    ));
+
+    assertThat(aggregationResult.get("count"), equalTo(45L));
```
kolea2, line 33: I would use `Truth.assertThat...` to be consistent with other tests.

kolea2 (changes requested):
looking really good!

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -337,33 +360,11 @@ public Iterator<Entity> get(Iterable<Key> keys, ReadOption... options) {
...
-      if (optionsByType.containsKey(ReadTime.class)) {
-        readOptionsPb =
-            com.google.datastore.v1.ReadOptions.newBuilder()
-                .setReadTime(((ReadTime) optionsByType.get(ReadTime.class)).time().toProto())
-                .build();
-      }
+      return this.readOptionProtoPreparer.prepare(Arrays.asList(options));
     }
-    return readOptionsPb;
+    return null;
```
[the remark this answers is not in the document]
jainsahab: Hi @kolea2 , I've maintained the existing contract of this method, that is to return null when no `ReadOptions` are provided by the user.
Since  we are using Java 8, how about if we use `Optional` here ? seems like a perfect use case.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  AggregationQuery(String namespace, GqlQuery<?> gqlQuery) {
+    super(namespace);
+    this.nestedGqlQuery = gqlQuery;
+    this.mode = GQL;
+  }
+
+  /** Returns the {@link Aggregation}(s) for this Query. */
+  public Set<Aggregation> getAggregations() {
```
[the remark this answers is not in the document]
jainsahab: https://github.com/googleapis/java-datastore/pull/823/#discussion_r972614041

Few things like multiple aggregations with same aliases can be solved by using set, and if we pass duplicate aggregation to backend it'll throw errors like:
image
By using a`Set` we can handle this mistake by the user and return the results gracefully.

Proto defines this field with `repeated` keyword. Though order doesn't matter in this case, As the enduser  gets the map (`Map<Alias, AggregationValue>`) in result. If we want to preserve the order of aggregations in aggregation query, we can use `LinkedHashSet`. Let me know what you think.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Query.java
```diff
@@ -156,26 +156,18 @@ static ResultType<?> fromPb(com.google.datastore.v1.EntityResult.ResultType type
...
-    return resultType;
-  }
-
   public String getNamespace() {
     return namespace;
   }
 
   ToStringHelper toStringHelper() {
-    return MoreObjects.toStringHelper(this).add("type", resultType).add("namespace", namespace);
+    return MoreObjects.toStringHelper(this).add("namespace", namespace);
```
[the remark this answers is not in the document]
jainsahab: `type` here represents ENTITY, PROJECTIONS or KEY. we have removed the coupling of such things from `Query`  and made it generic. Hence, this is removed from here and moved down in the hierarchy. ex: `StructuredQuery#toString()`

.gitignore
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/8c81b9f201102b4db79d195263d15fb06e4985d3.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
```diff
@@ -461,4 +461,54 @@ public interface Datastore extends Service<DatastoreOptions>, DatastoreReaderWri
...
+
+  /**
+   * Submits a {@link AggregationQuery} and returns {@link AggregationResults}.
+   * {@link ReadOption}s can be specified if desired.
+   *
+   * <p>Example of running an {@link AggregationQuery} to find the count of entities of one kind.
+   *
+   * <p>{@link StructuredQuery} example:</p>
+   * <pre>{@code
+   * import static com.google.cloud.datastore.aggregation.Aggregation.count;
```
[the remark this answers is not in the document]
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/8c81b9f201102b4db79d195263d15fb06e4985d3.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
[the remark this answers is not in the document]
jainsahab: Replaced the code dealing with null here with `Optional` here https://github.com/googleapis/java-datastore/pull/823/commits/1584d9f88ae7af6a87223daaf5d0af6690e5a017.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/AggregationResultTest.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/bc425ddc2a887020beb6fd49355ee96ac94a76b4.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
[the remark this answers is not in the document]
kolea2: great!

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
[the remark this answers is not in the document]
kolea2: got it - as long as it's doing what the backend expects, sounds good to me!

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResults.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * The result of an {@link AggregationQuery} query submission. Contains a {@link
+ * List<AggregationResult>} and readTime {@link Timestamp} in it.
+ *
+ * <p>This can be used to iterate over underlying {@link List<AggregationResult>} directly.
+ *
+ * <p>Though {@link com.google.cloud.datastore.aggregation.CountAggregation} is guaranteed to return
+ * only one {@link AggregationResult} as part of its execution.
+ *
+ * <p>In the future, we might support more complex {@link AggregationQuery} that might result in
```
kolea2, line 36: nit - I would remove this to keep the docs just about the class functionality itself

google-cloud-datastore/src/main/java/com/google/cloud/datastore/RetryAndTraceDatastoreRpcDecorator.java
```diff
@@ -0,0 +1,124 @@
...
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Status;
+import java.util.concurrent.Callable;
+
+/**
+ * An implementation of {@link DatastoreRpc} which acts as a Decorator and decorates the underlying
+ * {@link DatastoreRpc} with the logic of retry and Traceability.
+ */
+@InternalApi
+public class RetryAndTraceDatastoreRpcDecorator implements DatastoreRpc {
```
kolea2, line 52: because this is only implemented for aggregation queries, perhaps this name should be something different?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQuery.java
```diff
@@ -1018,6 +1019,8 @@ public Integer getLimit() {
 
   public abstract Builder<V> toBuilder();
 
+  @InternalApi
```
kolea2, line 1022: I don't think we should change this to internal

google-cloud-datastore/src/main/java/com/google/cloud/datastore/TransactionImpl.java
```diff
@@ -16,6 +16,8 @@
 
 package com.google.cloud.datastore;
 
+import static com.google.cloud.datastore.ReadOption.transactionId;
```
kolea2, line 19: I would be inclined to revert this, as it makes the readability more confusing (`transactionId(transactionId)`)

google-cloud-datastore/src/main/java/com/google/cloud/datastore/execution/response/AggregationQueryResponseTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+    implements ResponseTransformer<RunAggregationQueryResponse, AggregationResults> {
+
+  @Override
+  public AggregationResults transform(RunAggregationQueryResponse response) {
+    Timestamp readTime = Timestamp.fromProto(response.getBatch().getReadTime());
+    List<AggregationResult> aggregationResults =
+        response.getBatch().getAggregationResultsList().stream()
+            .map(
+                aggregationResult -> new AggregationResult(resultWithLongValues(aggregationResult)))
+            .collect(Collectors.toCollection(LinkedList::new));
```
kolea2, line 44: why a linked list here?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/ProtoTestData.java
```diff
@@ -63,6 +66,17 @@ public static PropertyReference propertyReference(String value) {
...
   }
 
+  public static Aggregation countAggregation(String alias) {
+    return Aggregation.newBuilder().setAlias(alias).setCount(Count.newBuilder().build()).build();
+  }
+
+  public static Aggregation countAggregation(String alias, long limit) {
+    return Aggregation.newBuilder()
+        .setAlias(alias)
+        .setCount(Count.newBuilder().setUpTo(Int64Value.of(limit)))
```
kolea2, line 76: I don't think we want to use up to? Rather, can we use limit on the underlying query?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/execution/AggregationQueryExecutorTest.java
```diff
@@ -0,0 +1,177 @@
...
+                    new AggregationResult(
+                        ImmutableMap.of(
+                            "count", LongValue.of(209), "property_2", LongValue.of(100))),
+                    new AggregationResult(
+                        ImmutableMap.of(
+                            "count", LongValue.of(509), "property_2", LongValue.of(100)))),
+                Timestamp.fromProto(runAggregationQueryResponse.getBatch().getReadTime())));
+  }
+
+  private RunAggregationQueryResponse dummyAggregationQueryResponse() {
```
kolea2, line 146: rename to `placeholderAggregationQueryResponse`

google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreTest.java
```diff
@@ -1311,4 +1344,28 @@ public void testQueryWithStartCursor() {
     assertEquals(cursor2, cursor1);
     datastore.delete(entity1.getKey(), entity2.getKey(), entity3.getKey());
   }
+
+  private RunAggregationQueryResponse dummyAggregationQueryResponse() {
```
kolea2, line 1348: rename to `placeholderAggregationQueryResponse`

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/TransactionImpl.java
jainsahab: how about we change the method name to `withTransaction`? In that case it becomes `withTransaction(transactionId)`.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/execution/response/AggregationQueryResponseTransformer.java
jainsahab: while writing this i felt LinkedList will be more performant here because:
 - If we have too many `AggregationResult` coming in, then `ArrayList` will need to double it's capacity everytime it's full and copy the existing data in it whereas in `LinkedList` it'll be a O(1) operation.
 - We are exposing the result as an iterator anyway and won't need a case where we need to access the result using random index where `ArrayList` is optimized.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreTest.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/a5b462332db2f8b88db4016d2f83046c89e74acd.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/execution/AggregationQueryExecutorTest.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/pull/823/commits/a5b462332db2f8b88db4016d2f83046c89e74acd.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResults.java
jainsahab: Fixed as part of https://github.com/googleapis/java-datastore/commit/a5b462332db2f8b88db4016d2f83046c89e74acd.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/TransactionImpl.java
kolea2: that sounds good!

kolea2: or `setTransactionId` also sgtm

gforgurups (approved):

kolea2 (approved):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/TransactionImpl.java
jainsahab: addressed in https://github.com/googleapis/java-datastore/pull/1041.

PR #875, opened 2022-10-13
refactor: making query class generic 

jainsahab:
This change decouples Query.java from the concerns related to handling entity records. It is required as the definition of a Query is changed now and it is not just limited to the the fact that every query returns entity records, With the introduction of count feature and aggregation query, we will have queries returning aggregation results.

_Note: This PR is intended to trim down the changes done as part of #823._

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/RecordQuery.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/RecordQuery.java
@@ -22,6 +22,7 @@ import com.google.cloud.datastore.Query.ResultType;
 @InternalApi
 public interface RecordQuery<V> {
 
+  @InternalApi
   ResultType<V> getType();
 
   @InternalApi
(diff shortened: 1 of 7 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/RecordQuery.java
```diff
@@ -0,0 +1,32 @@
...
+package com.google.cloud.datastore;
+
+import com.google.api.core.InternalApi;
+import com.google.cloud.datastore.Query.ResultType;
+
+/** An internal marker interface to represent {@link Query} that returns the entity records. */
+@InternalApi
+public interface RecordQuery<V> {
+
+  ResultType<V> getType();
```
kolea2, line 25: is this also meant to be internal?

kolea2 (approved):
lgtm with one outstanding question

google-cloud-datastore/src/main/java/com/google/cloud/datastore/RecordQuery.java
jainsahab: Yup, it was package private earlier, so an internal function here also. just fixed in https://github.com/googleapis/java-datastore/pull/875/commits/aa834031e36be75c489525fca1d4734edf374ba2.

PR #876, opened 2022-10-13
refactor: Extracting out Proto preparation logic to encourage reusability

jainsahab:
This change introduces a component called ProtoPreparer which will be used to populate the protos from the domain objects. In the upcoming count aggregation feature, aggregation query will be acting upon StructuredQuery and GqlQuery and hence need to populate the protos of those queries **again**.  For that reason _ProtoPreparer_ component will have the responsibility to populate the protos and be reused in multiple usecases.

_Note: This PR is intended to trim down the changes done as part of #823 ._


google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
```diff
@@ -452,6 +452,14 @@ public Map<String, Object> getNamedBindings() {
     return builder.buildOrThrow();
   }
 
+  public Map<String, Binding> getNamedBindingsMap() {
```
kolea2, line 455: we might want to consider reducing this to `protected` or marking as InternalApi (same for positional bindings)

google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOptionProtoPreparer.java
```diff
@@ -0,0 +1,77 @@
+/*
```
kolea2, line 1: This is the only preparer that is a bit confusing to me - where is this logic extracted from?

If this is not a refactor, we should do this as a feature separately. This PR should focus only on the refactoring of code.

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOptionProtoPreparer.java
jainsahab: This is extracted from DatastoreImpl.java [this location to be precise](https://github.com/googleapis/java-datastore/pull/876/files#diff-f6b1d8b631b8df3355d5820957585afd4ae2fb5df43357e10dc29828746c2fdfL340-L367).

It's an old code which is extracted to this ProtoPreparer, so that it can be reused in AggregationQuery in count feature.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
jainsahab: fixed as part of https://github.com/googleapis/java-datastore/pull/876/commits/28e0332606128204ba3590c9f13007bbbf2cb555.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/ReadOptionProtoPreparer.java
kolea2: Got it. I still see some inconsistencies - can you confirm these are intended?

1. https://github.com/googleapis/java-datastore/pull/876/files#diff-f6b1d8b631b8df3355d5820957585afd4ae2fb5df43357e10dc29828746c2fdfL346-L350. This check appears to be missing
2. https://github.com/googleapis/java-datastore/pull/876/files#diff-1a25d0e25821474babf2fe4712c0895b18dbac81f3b0883581cb58bf601693d2R59-R65 This seems new

kolea2: Also https://github.com/googleapis/java-datastore/pull/876/files#diff-1a25d0e25821474babf2fe4712c0895b18dbac81f3b0883581cb58bf601693d2R42-R46 seems new

kolea2: Taking a closer look, I see what the intention is by first checking for more than one key. What happens if you set both eventual consistency and transaction id?

kolea2: Discussed offline, readoptions can only have one consistency type; see https://github.com/googleapis/java-datastore/blob/main/proto-google-cloud-datastore-v1/src/main/proto/google/datastore/v1/datastore.proto#L462-L499

kolea2 (approved):

PR #923, opened 2022-11-21
fix: #355 Explicitly passing --project argument when starting emulator

### LINKED ISSUE #355: Datastore emulator fails to start without project id property
#### Environment details

1. GCloud datastore emulator 2.1.0
2. Google Cloud SDK 327.0.0
2. OS type and version: NAME="Ubuntu" VERSION="18.04.4 LTS (Bionic Beaver)"  / MacOS Catalina 10.15.7 (19H15)
3. Java version: OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.8+10) / OpenJDK Runtime Environment (build 11.0.7+10)
4. Libs:
* https://github.com/GoogleCloudPlatform/spring-cloud-gcp: `org.springframework.cloud:spring-cloud-gcp-starter-data-datastore:1.2.5.RELEASE`, I checked - same with the latest
* `google-cloud-datastore:1.104.0`
* `google-cloud-core:1.93.7`

#### Steps to reproduce
This version emulator fails to start without `--project` property

  1. `gcloud beta emulators datastore start`

```sh
ERROR: (gcloud.beta.emulators.datastore.start) The required property [project] is not currently set.
You may set it for your current workspace by running:

  $ gcloud config set project VALUE

or it can be set temporarily by the environment variable [CLOUDSDK_CORE_PROJECT]
```

Starting emulator with Java code like this:

```java
LocalDatastoreHelper.create(settings.getConsistency(), settings.getPort());
```

Causes: no messages in logs, the application failed to call the Datastore then with I/O exceptions, i.e. curl doesn't work.

### Workaround

Set env variable: `CLOUDSDK_CORE_PROJECT` with the projectId value

### Questions

1. Why `LocalDatastoreHelper` doesn't pass `--project` on startup and doesn't expect this from the builder?
2. Shouldn't `LocalDatastoreHelper #start` fail to start on this kind of errors?

Besides the outcomes of these questions, I would expect this class to fail to start or log something.

jainsahab:
fix #355 

**Context:**
LocalDatastoreHelper starts a subprocess `gcloud beta emulators datastore start` to start the emulator, It is always starting the subprocess under the assumption that user have already set the project id property using `gcloud config set project some-project-id`.  If the `project-id` property is set, it'll result in successful execution of sub-process. Otherwise it'll throw the following error.

```
➜  java-datastore git:(main) gcloud beta emulators datastore start
WARNING: Reusing existing data in [/Users/jainsahab/.config/gcloud/emulators/datastore].
Executing: /Users/jainsahab/Work/bin/google-cloud-sdk/platform/cloud-datastore-emulator/cloud_datastore_emulator start --host=::1 --port=8045 --store_on_disk=True --allow_remote_shutdown --consistency=0.9 /Users/jainsahab/.config/gcloud/emulators/datastore
ERROR: (gcloud.beta.emulators.datastore.start) The required property [project] is not currently set.
It can be set on a per-command basis by re-running your command with the [--project] flag.

You may set it for your current workspace by running:

  $ gcloud config set project VALUE

or it can be set temporarily by the environment variable [CLOUDSDK_CORE_PROJECT]
```
_Note: To unset the project-id property, run `gcloud config unset project`_

[BlockingProcessStreamReader.java](https://github.com/googleapis/java-core/blob/main/google-cloud-core/src/main/java/com/google/cloud/testing/BlockingProcessStreamReader.java#L47-L61) file from [google-cloud-core library](https://github.com/googleapis/java-core) which is a java Thread **ignores this helpful text and the subprocess's exit code** to silently completes its own execution. Hence, no error was given to the user.

**Solution:**
As part of this PR, [google-cloud-core library](https://github.com/googleapis/java-core) (_[covered in this PR](https://github.com/googleapis/java-core/pull/1039)_) is out of scope, but we are explicitly passing [--project argument](https://cloud.google.com/sdk/gcloud/reference#--project) now while starting the emulator so that it no longer expect the `project-id` property to set.
In addition to that, we are also giving the user the capability to set a custom project Id.

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -140,6 +140,7 @@ public class LocalDatastoreHelper extends BaseEmulatorHelper<DatastoreOptions> {
         "datastore",
         builder.port > 0 ? builder.port : BaseEmulatorHelper.findAvailablePort(DEFAULT_PORT),
         firstNonNull(builder.projectId, DEFAULT_PROJECT_ID));
+    String projectId = firstNonNull(builder.projectId, DEFAULT_PROJECT_ID);
     this.consistency = builder.consistency > 0 ? builder.consistency : DEFAULT_CONSISTENCY;
     this.gcdPath = builder.dataDir;
     this.storeOnDisk = builder.storeOnDisk;
@@ -150,7 +151,7 @@ public class LocalDatastoreHelper extends BaseEmulatorHelper<DatastoreOptions> {
     List<String> gcloudCommand = new ArrayList<>(Arrays.asList(GCLOUD_CMD_TEXT.split(" ")));
     gcloudCommand.add(GCLOUD_CMD_PORT_FLAG + "localhost:" + getPort());
     gcloudCommand.add(CONSISTENCY_FLAG + builder.consistency);
-    gcloudCommand.add(PROJECT_FLAG + getProjectId());
+    gcloudCommand.add(PROJECT_FLAG + projectId);
     if (!builder.storeOnDisk) {
       gcloudCommand.add("--no-store-on-disk");
     }
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -68,6 +68,7 @@ public class LocalDatastoreHelper extends BaseEmulatorHelper<DatastoreOptions> {
 
   // Common settings
   private static final String CONSISTENCY_FLAG = "--consistency=";
+  private static final String PROJECT_FLAG = "--project=";
   private static final double DEFAULT_CONSISTENCY = 0.9;
 
   private static final Logger LOGGER = Logger.getLogger(LocalDatastoreHelper.class.getName());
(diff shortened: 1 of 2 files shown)


jainsahab:
PR for the issue in google-cloud-core library is up for review too @ https://github.com/googleapis/java-core/pull/1039.

artemptushkin (approved):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
```diff
@@ -140,6 +150,7 @@ private LocalDatastoreHelper(Builder builder) {
     List<String> gcloudCommand = new ArrayList<>(Arrays.asList(GCLOUD_CMD_TEXT.split(" ")));
     gcloudCommand.add(GCLOUD_CMD_PORT_FLAG + "localhost:" + getPort());
     gcloudCommand.add(CONSISTENCY_FLAG + builder.consistency);
+    gcloudCommand.add(PROJECT_FLAG + getProjectId());
```
kolea2, line 153: I think this should only be set if the project id is not null

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
jainsahab: Yeah, it seems confusing at first, but `getProjectId` will always return a non null value in this case, because we are passing a non null value to the `projectId` parameter in super constructor line 142 above..

jainsahab: I've created a local variable now to eradicate the confusion.

kolea2 (approved):

PR #940, opened 2022-12-16
feat: add x-goog-request-params to header

kolea2:


datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java
```diff
@@ -123,8 +128,16 @@ public InputStream call(String methodName, MessageLite request) throws Datastore
...
   @VisibleForTesting
-  void setHeaders(MessageLite request, HttpRequest httpRequest) {
+  void setHeaders(
+      MessageLite request, HttpRequest httpRequest, String projectId, String databaseId) {
     httpRequest.getHeaders().put(API_FORMAT_VERSION_HEADER, API_FORMAT_VERSION);
+    StringBuilder builder = new StringBuilder("project_id=");
+    builder.append(projectId);
+    if (!Strings.isNullOrEmpty(databaseId)) {
+      builder.append("&database_id=");
+      builder.append(databaseId);
```
viacheslav-rostovtsev, line 138: This will never contain anything that needs to be encoded, right? Database names are (ALPHA / DIGIT / "-" / "." / "_" / "~")?

viacheslav-rostovtsev (approved):
LGTM, but check the comment

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java
```diff
@@ -46,6 +47,8 @@ class RemoteRpc {
   @VisibleForTesting static final String API_FORMAT_VERSION_HEADER = "X-Goog-Api-Format-Version";
   private static final String API_FORMAT_VERSION = "2";
 
+  @VisibleForTesting static final String X_GOOG_REQUEST_PARAMS_HEADER = "x-goog-request-params";
```
blakeli0, line 50: This makes me think that we should expose this key as public in gax-java in the future, we have the same key defined [here](https://github.com/googleapis/gax-java/blob/a1d62706624539a699745a22241c9ebc14b3495d/gax-grpc/src/main/java/com/google/api/gax/grpc/CallOptionsUtil.java#L48)

blakeli0:
I know we probably only need `databaseId` and `projectId` at this moment, but in case we need more info that we cannot easily get from the request directly in the future, do we want to implement extract and match logic based on the proto? So that we don't need to change the code every time the routing annotation changes. 
We implemented it for pure gRPC gapic client libraries, and [this](https://github.com/googleapis/gax-java/blob/a1d62706624539a699745a22241c9ebc14b3495d/gax-grpc/src/main/java/com/google/api/gax/grpc/CallOptionsUtil.java#L48) is how the autogenerated client for Bigtable look like(I know Bigtable does not really use it, but you get the idea).

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java
kolea2: @blakeli0 makes sense! This module doesn't depend on gax, but let me know if you'd like any similar changes here.

kolea2:
> I know we probably only need `databaseId` and `projectId` at this moment, but in case we need more info that we cannot easily get from the request directly in the future, do we want to implement extract and match logic based on the proto? So that we don't need to change the code every time the routing annotation changes. We implemented it for pure gRPC gapic client libraries, and [this](https://github.com/googleapis/gax-java/blob/a1d62706624539a699745a22241c9ebc14b3495d/gax-grpc/src/main/java/com/google/api/gax/grpc/CallOptionsUtil.java#L48) is how the autogenerated client for Bigtable look like(I know Bigtable does not really use it, but you get the idea).

@blakeli0 this is an interesting idea - I will file an issue to look into this as a refactor for the future

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java
kolea2: Yes, that's correct. See: https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases/create#query-parameters

blakeli0 (approved):

triplequark (approved):

PR #1032, opened 2023-03-24
docs: adds OR filter sample

telpirion:


samples/snippets/src/test/java/com/example/datastore/OrFilterQuerySampleIT.java
```diff
@@ -0,0 +1,63 @@
...
+  private ByteArrayOutputStream bout;
+  private PrintStream out;
+  private PrintStream originalPrintStream;
+
+  @Before
+  public void setUp() {
+    bout = new ByteArrayOutputStream();
+    out = new PrintStream(bout);
+    originalPrintStream = System.out;
+    System.setOut(out);
```
jainsahab, line 45: We can use `com.rule.SystemsOutRule` to avoid this boilerplate code, `com.example.datastore.AggregationQuerySampleTestIT` also use this junit rule which we can use as a reference.

jainsahab (changes requested):

samples/snippets/src/main/java/com/example/datastore/filters/OrFilterQuery.java
```diff
@@ -0,0 +1,61 @@
...
+    // Build the query
+    Query<Entity> query = Query.newEntityQueryBuilder()
+        .setKind("Task")
+        .setFilter(orFilter)
+        .build();
+
+    // Get the results back from Datastore
+    QueryResults<Entity> results = datastore.run(query);
+
+    if (results.hasNext()) {
```
kolea2, line 55: maybe throw an exception in the else since we always expect something to be returned here

samples/snippets/src/test/java/com/example/datastore/OrFilterQuerySampleIT.java
telpirion: Done.

samples/snippets/src/main/java/com/example/datastore/filters/OrFilterQuery.java
telpirion: Changed. I had intended to put `while` there but accidentally put `if`. Nice catch!

*Note*: I'm assuming that this is the correct way to iterate through a list of Datastore entities in Java. Please let me know if there is a better way.

samples/snippets/src/main/java/com/example/datastore/filters/OrFilterQuery.java
```diff
@@ -0,0 +1,61 @@
...
+    // Build the query
+    Query<Entity> query = Query.newEntityQueryBuilder()
+        .setKind("Task")
+        .setFilter(orFilter)
+        .build();
+
+    // Get the results back from Datastore
+    QueryResults<Entity> results = datastore.run(query);
+
+    while (results.hasNext()) {
```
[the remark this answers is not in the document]
telpirion: Done!

kolea2 (commented):
current changes look good, it would be nice to also have a nested OR example. Will defer to @jlara310 here for requirements

telpirion:
@kolea2  I think we should discuss a nested OR query in a follow-up PR. WDYT?

jainsahab (approved):

PR #1041, opened 2023-04-04
chore: renaming transactionId to setTransactionId to increase readability

jainsahab:
addressing https://github.com/googleapis/java-datastore/pull/823#discussion_r995924998


kolea2:
Looking at this again, I didn't notice the other methods in the class are also missing the `set...` prefix. I suppose it makes more sense to be consistent. Sorry for the back and forth, I'll close this out and the issue.

PR #1067, opened 2023-05-08
feat: Sum and Avg aggregation feature

jainsahab:
Add support for querying the mathematical SUM and AVERAGE of an entity's specific field chosen by user.


google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
```diff
@@ -143,6 +144,18 @@ public Builder addAggregation(Aggregation aggregation) {
...
     }
 
+    public Builder addAggregations(AggregationBuilder<?>... aggregationBuilders) {
+      for (AggregationBuilder<?> builder : aggregationBuilders) {
+        this.aggregations.add(builder.build());
+      }
+      return this;
+    }
+
+    public Builder addAggregations(Aggregation... aggregations) {
```
kolea2, line 154: maybe add one more convenience method accepting a List?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
```diff
@@ -38,7 +41,19 @@ public AggregationResult(Map<String, LongValue> properties) {
    * @return An aggregation result value for the given alias.
    */
   public Long get(String alias) {
```
kolea2, line 43: I would also add getLong, and we can encourage users to use the explicit type

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
```diff
@@ -0,0 +1,72 @@
+package com.google.cloud.datastore.aggregation;
```
kolea2, line 1: missing header

google-cloud-datastore/src/test/java/com/google/cloud/datastore/aggregation/AvgAggregationTest.java
```diff
@@ -0,0 +1,72 @@
+package com.google.cloud.datastore.aggregation;
```
kolea2, line 1: missing header

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
jainsahab: sure, do you think it would be good idea to mark older method `get` with @Deprecated annotation ?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/SumAggregation.java
```diff
@@ -0,0 +1,72 @@
...
+    return aggregationBuilder().setSum(sum).build();
+  }
+
+  @Override
+  protected boolean sameAs(Aggregation aggregation) {
+    SumAggregation that = (SumAggregation) aggregation;
+    if (!this.propertyReference.equals(that.propertyReference)) {
+      return false;
+    }
+    boolean bothAliasAreNull = getAlias() == null && that.getAlias() == null;
```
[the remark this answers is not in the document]
jainsahab: property reference field is limited to sum/avg aggregation only, abstract class doesn't have any knowledge of that field, If we want to involve that into equality logic, we will have to keep the logic in this class only.

Or we can move the `alias` equality logic to super class, and call it from every children classes.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/execution/response/AggregationQueryResponseTransformerTest.java
```diff
@@ -45,13 +47,13 @@ public void shouldTransformAggregationQueryResponse() {
...
                 "count", intValue(209),
-                "property_2", intValue(100)));
+                "property_2", doubleValue(100)));
 
     Map<String, com.google.datastore.v1.Value> result2 =
         new HashMap<>(
             ImmutableMap.of(
                 "count", intValue(509),
-                "property_2", intValue(100)));
+                "property_2", doubleValue(100)));
```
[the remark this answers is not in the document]
jainsahab: With the introduction of SUM/AVG aggregation, we might have value of double type in response and that's what we are trying to simulate here.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
jainsahab: fixed as part of 403c01c

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
jainsahab: fixed as part of 403c01c.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
jainsahab: fixed as part of 403c01c.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
```diff
@@ -0,0 +1,72 @@
...
+  private final String propertyReference;
+
+  public AvgAggregation(String alias, String propertyReference) {
+    super(alias);
+    checkArgument(propertyReference != null, "Property reference can't be null");
+    this.propertyReference = propertyReference;
+  }
+
+  @Override
+  public AggregationQuery.Aggregation toPb() {
```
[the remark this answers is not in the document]
jainsahab: fixed as part of 403c01c.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
```diff
@@ -0,0 +1,72 @@
...
+    return aggregationBuilder().setAvg(avg).build();
+  }
+
+  @Override
+  protected boolean sameAs(Aggregation aggregation) {
+    AvgAggregation that = (AvgAggregation) aggregation;
+    if (!this.propertyReference.equals(that.propertyReference)) {
+      return false;
+    }
+    boolean bothAliasAreNull = getAlias() == null && that.getAlias() == null;
```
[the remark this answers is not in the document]
jainsahab: Thanks for pointing this out, fixed as part of 403c01c.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
```diff
@@ -0,0 +1,72 @@
...
+      return true;
+    } else {
+      boolean bothArePresent = getAlias() != null && that.getAlias() != null;
+      return bothArePresent && getAlias().equals(that.getAlias());
+    }
+  }
+
+  @Override
+  public int hash() {
+    return Objects.hash(getAlias());
```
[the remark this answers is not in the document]
jainsahab: yes, `hash` is being invoked from super classes's hashcode only.

To give you some context, I overridden `equals` method to move [some boiler plate code](https://github.com/jainsahab/java-datastore/blob/403c01c8fcdb7be918984f3c5d32bfc92aeb7b6d/google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/Aggregation.java#L40-L47) in a single place and keep the concrete `Aggregation` class clean as much as possible, 


ErrorProne started giving error about the hashcode now, Apparently if a class implements equals, it must implemtn hashCode too.
image

even if the children classes implements the hashcode, it won't budge and won't let it compile.

So finally, I moved `equals` and `hashcode` to the abstract class and kept the boilerplate code in a single place, 
and introduced two equivalent methods `sameAs` and `hash` in the children classes.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/SumAggregation.java
```diff
@@ -0,0 +1,72 @@
+package com.google.cloud.datastore.aggregation;
```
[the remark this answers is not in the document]
jainsahab: fixed as part of 403c01c.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/aggregation/SumAggregationTest.java
```diff
@@ -0,0 +1,74 @@
+package com.google.cloud.datastore.aggregation;
```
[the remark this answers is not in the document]
jainsahab: fixed as part of 403c01c.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreAggregationsTest.java
```diff
@@ -0,0 +1,145 @@
+package com.google.cloud.datastore.it;
```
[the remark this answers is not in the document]
jainsahab: fixed as part of 403c01c.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/execution/response/AggregationQueryResponseTransformerTest.java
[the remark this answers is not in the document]
kolea2: let's instead keep these tests as is and add new ones speciically for double values

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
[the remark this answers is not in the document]
kolea2: What happens if all the children implement equals and hashCode concretely? It looks like by doing it in the above way, CLIRR is breaking which we don't want.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/execution/response/AggregationQueryResponseTransformerTest.java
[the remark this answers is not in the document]
jainsahab: sure, fixed as part of fc666ba.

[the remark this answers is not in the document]
jainsahab: sure, fixed as part of fc666ba.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/AvgAggregation.java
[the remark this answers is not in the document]
jainsahab: It's just some boilderplate code of `equals` getting replicated in all Aggregation implementations, sure, fixed as part of fc666ba.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
kolea2: sorry, I meant adding something like: 

`public Builder addAggregations(List<Aggregation> aggregations)`

if you think this is not a useful convenience method, feel free to ignore :)

kolea2 (commented):
looking really good!

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationQuery.java
jainsahab: sure, Let's keep and expose the minimal API for now.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreAggregationsTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+  @Test
+  public void testAggregationQueryWithSumAggregation() {
+    DATASTORE.put(entity1, entity2);
+
+    EntityQuery baseQuery = Query.newEntityQueryBuilder().setKind(KIND).build();
+    AggregationQuery aggregationQuery =
+        Query.newAggregationQueryBuilder()
+            .over(baseQuery)
+            .addAggregations(sum("marks").as("total_marks"))
```
kolea2, line 74: can we add one test without an alias as well?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/aggregation/Aggregation.java
```diff
@@ -38,8 +39,30 @@ public String getAlias() {
...
+    return aggregationBuilder;
+  }
+
   /** Returns a {@link CountAggregation} builder. */
   public static CountAggregation.Builder count() {
     return new CountAggregation.Builder();
   }
+
+  /** Returns a {@link SumAggregation} builder. */
+  @BetaApi
```
kolea2, line 58: please remove (and elsewhere)

jainsahab: thanks for pointing it out, [c590fb4](https://github.com/googleapis/java-datastore/pull/1067/commits/c590fb44ba0341627bcad5d640c4586aa6889f8b) addresses it.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreAggregationsTest.java
jainsahab: added tests for both sum/avg in [30c0de3](https://github.com/googleapis/java-datastore/pull/1067/commits/30c0de3679a67535739b52df96508b029c93f494).

kolea2 (approved):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
```diff
@@ -24,21 +27,52 @@
...
+   * @param alias A custom alias provided in the query or an autogenerated alias in the form of
+   *     'property_\d'
+   * @return An aggregation result value for the given alias.
+   */
+  public Long getLong(String alias) {
+    Value<?> value = properties.get(alias);
+    if (value.getType() == DOUBLE) {
+      return ((Double) value.get()).longValue();
+    }
+    return (Long) value.get();
```
meltsufin, line 60: This may not be a safe cast if more types are supported in the future.
Maybe check for `== LONG` here.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
```diff
@@ -24,21 +27,52 @@
...
+   * @param alias A custom alias provided in the query or an autogenerated alias in the form of
+   *     'property_\d'
+   * @return An aggregation result value for the given alias.
+   */
+  public Double getDouble(String alias) {
+    Value<?> value = properties.get(alias);
+    if (value.getType() == LONG) {
+      return ((Long) value.get()).doubleValue();
+    }
+    return (Double) value.get();
```
meltsufin, line 75: Similarly here, maybe check for `== DOUBLE`.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
```diff
@@ -24,21 +27,52 @@
 /** Represents a result of an {@link AggregationQuery} query submission. */
 public class AggregationResult {
 
-  private final Map<String, LongValue> properties;
+  private final Map<String, Value<?>> properties;
 
-  public AggregationResult(Map<String, LongValue> properties) {
+  public AggregationResult(Map<String, Value<?>> properties) {
```
meltsufin, line 32: Are we certain that this is not a breaking change?
Maybe to be safe, just add another constructor
`public AggregationResult(Map<String, DoubleValue> properties)`?
cc: @burkedavison for a 2nd opinion.

burkedavison: Re: another constructor
You can't have both `public AggregationResult(Map<A, B> map)` and `public AggregationResult(Map<A, C> map)`, as their signatures overlap (`public AggregationResult(Map map)` once type erasure occurs. It will throw a compiler error if you give it a try.

Re: breaking change
```java
Map<String, LongValue> longMap = Collections.emptyMap();
new AggregationResult(longMap); // This will also create a compiler error with the current syntax.
```
Instead, use:
```java
public AggregationResult(Map<String, ? extends Value<?>> properties)
```
and then the compiler will be happy.

meltsufin: Thanks!

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
jainsahab: addressed as part of [b60995b](https://github.com/googleapis/java-datastore/pull/1067/commits/b60995b0be6813d5f302006b02ae74729d0aa33f).

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
jainsahab: adressed as part of [b60995b](https://github.com/googleapis/java-datastore/pull/1067/commits/b60995b0be6813d5f302006b02ae74729d0aa33f).

google-cloud-datastore/src/main/java/com/google/cloud/datastore/AggregationResult.java
jainsahab: Thanks for pointing it out, have addressed this in [b60995b](https://github.com/googleapis/java-datastore/pull/1067/commits/b60995b0be6813d5f302006b02ae74729d0aa33f).

meltsufin (approved):

PR #1082, opened 2023-05-24
docs: Marking Transaction, Batch and DatastoreBatchWriter class with 'NotThreadSafe' annotation

### LINKED ISSUE #295: Transaction "add"s are not multi-thread safe
The `add(entity)` method in a `Transaction` object is not multi-thread safe. Other methods (e.g. put, delete) may also not be safe but we have traced a failure in our server specifically to `add`. Our server uses Scala Futures and so uses multiple threads of execution. 

The particular scenario we tracked down was in our new user creation processing where we add 4 entities in a transaction. However there was a probability that one of the entities didn't get saved. It wasn't always the same entity kind that "disappeared". We traced this by adding debug logging after the 'add' calls that shows the `add`s successfully completing - but subsequent processing would find one of the entities missing. We did note that it always seemed to be just one entity missing and when an entity did go missing it seemed to be the first entity kind to be added (thread execution order was not consistent across invocations).

We added `synchronize` on each of our DataStore calls and haven't encountered this issue since. So at a minimum this issue can serve as a warning to others. Maybe a note could be added to the documentation. But we are also patiently waiting for issue #22 (async support) to be implemented and would hope that explicit synchronization to not be needed.

jainsahab:
Documenting the thread unsafe behaviour of Transaction, Batch and DatastoreBatchWriter classes.

Fixes #295

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Batch.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Batch.java
@@ -34,12 +34,12 @@ import javax.annotation.concurrent.NotThreadSafe;
  * batch.submit();
  * }</pre>
  *
- * <p><b> WARNING: THIS CLASS MAINTAINS AN INTERNAL STATE IN TERMS OF {@link
- * java.util.LinkedHashMap} AND {@link java.util.LinkedHashSet} WHICH GETS UPDATED ON EVERY METHOD
- * CALL PERFORMING CRUD OPERATIONS TO RECORD THE MUTATIONS, SINCE {@link java.util.LinkedHashMap} IS
- * NOT THREAD SAFE AS PER ITS <a
- * href="https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html">DOCUMENTATION</a>.
- * THIS CLASS TOO SHOULD NOT BE TREATED AS A THREAD SAFE CLASS. </b>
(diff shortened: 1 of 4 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/Batch.java
```diff
@@ -32,7 +33,15 @@
  * batch.add(entity2, entity3);
  * batch.submit();
  * }</pre>
+ *
+ * <p><b> WARNING: THIS CLASS MAINTAINS AN INTERNAL STATE IN TERMS OF {@link
+ * java.util.LinkedHashMap} AND {@link java.util.LinkedHashSet} WHICH GETS UPDATED ON EVERY METHOD
+ * CALL PERFORMING CRUD OPERATIONS TO RECORD THE MUTATIONS, SINCE {@link java.util.LinkedHashMap} IS
```
kolea2, line 39: two nits -
1. I would use lowercasing here (maybe just keep "WARNING" in all caps?)
2. Run on sentence, I would rephrase: (`...to record the mutations. Since LinkedHashMap...`) (swap out the comma with a period)

kolea2 (approved):
lgtm with two small nits

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Batch.java
jainsahab: fixed in fb6c899.

PR #1083, opened 2023-05-24
test: use ITDatastoreConceptsTest for graalvm testing

### LINKED ISSUE #1070: Run graalvm tests with samples
Reconfigure graalvm presubmits to run against samples tests

kolea2:
Fixes #1070 

This test aims to achieve the same goal (to test basic functionality in GraalVM presubmits) as https://github.com/googleapis/java-storage/pull/1872. 

The test used is based largely on https://github.com/googleapis/java-datastore/blob/main/samples/snippets/src/test/java/com/google/datastore/snippets/ConceptsTest.java, but refactored to run only against the Datastore backend (rather than the emulator), and moved into google-cloud-datastore so we did not have to modify the templated run config.

### DIFF
--- a/google-cloud-datastore/src/test/resources/index.yaml
+++ b/google-cloud-datastore/src/test/resources/index.yaml
@@ -1,3 +1,5 @@
+# one time index creation is required to run ITDatastoreConceptsTest
+# see https://cloud.google.com/sdk/gcloud/reference/datastore/indexes/create for more details
 indexes:
   - kind: TaskConcepts
     properties:
--- a/google-cloud-datastore/src/test/resources/index.yaml
+++ b/google-cloud-datastore/src/test/resources/index.yaml
@@ -1,0 +1,32 @@
+indexes:
+  - kind: TaskConcepts
+    properties:
+      - name: done
+      - name: priority
+        direction: desc
+  - kind: TaskConcepts
+    properties:
+      - name: tag
+      - name: tag
+  - kind: TaskConcepts
+    properties:
+      - name: priority
+      - name: created
+  - kind: TaskConcepts
+    properties:
+      - name: category
+      - name: priority
+  - kind: TaskConcepts
+    properties:
+      - name: priority
+        direction: desc
+      - name: created
+  - kind: TaskConcepts
+    properties:
+      - name: percent_complete
+      - name: priority
+  - kind: TaskConcepts
+    properties:
+      - name: done
+      - name: priority
+        direction: desc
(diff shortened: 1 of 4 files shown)


google-cloud-datastore/src/test/resources/index.yaml
```diff
@@ -0,0 +1,32 @@
+indexes:
+  - kind: TaskConcepts
+    properties:
+      - name: done
+      - name: priority
+        direction: desc
+  - kind: TaskConcepts
+    properties:
+      - name: tag
+      - name: tag
```
jainsahab, line 10: IIUC, `tag` column should be listed only once.

jainsahab: this file is required for [composite index configurations](https://cloud.google.com/datastore/docs/tools/indexconfig), In that case should `tag` (only one column) even be mentioned here?

google-cloud-datastore/src/test/resources/index.yaml
```diff
@@ -0,0 +1,32 @@
+indexes:
```
jainsahab, line 1: How are we making use of this file? I couldn't find any script which deploy this file to datastore service.

google-cloud-datastore/src/test/resources/index.yaml
kolea2: When I run `testInQuerySorted` without the index, I get the following error:
```
com.google.cloud.datastore.DatastoreException: no matching index found. recommended index is:
- kind: TaskConcepts
  properties:
  - name: tag
  - name: tag
 ```

I **assume** this is because of the orderBy on the query, but not sure. Listing `tag` once gives an additional error (that it is a single property index), so I think this is the correct solution given by the error message.

google-cloud-datastore/src/test/resources/index.yaml
kolea2: Sorry for not explaining this earlier. As in the previous comment, some of these tests/queries require indexes to run. This is a one time setup done by running `gcloud datastore indexes create index.yaml`. While this does not need to be run in a script repeatedly, I thought it would be useful to include the index file as a resource for developers who want to run this test in the future so they don't need to figure out the indexes themselves.

jainsahab: Aah, that make sense, Can we add a comment on top of this file documenting this context ?

jainsahab (approved):
LGTM, just one nit. a comment on top of index.yaml describing its purpose is enough.

Issue #1054, opened 2023-04-26
AggregationQuery doesn't respect namespace from `com.google.cloud.datastore.Datastore` object


jainsahab:
#### Given
I have `2 entities` stored for `Task` kind under the namespace `ns1` and i configure a `com.google.cloud.datastore.Datastore` service with `ns1` namespace.

#### Scenario
When I try to run **entity** query with or without namespace through the `com.google.cloud.datastore.Datastore` object configured with namespace, i get same results, whereas in case of `AggregationQuery` i don't.

#### Code example

```java
import static com.google.cloud.datastore.aggregation.Aggregation.count;
import static com.google.common.collect.Iterables.getOnlyElement;

import com.google.cloud.datastore.AggregationQuery;
import com.google.cloud.datastore.Datastore;
import com.google.cloud.datastore.DatastoreOptions;
import com.google.cloud.datastore.Entity;
import com.google.cloud.datastore.EntityQuery;
import com.google.cloud.datastore.Query;
import java.util.ArrayList;
import java.util.Iterator;
import java.util.List;

public class Playground {

  public static void main(String[] args) {
    DatastoreOptions datastoreOptions = DatastoreOptions.newBuilder()
        .setNamespace("ns1")
        .build();
    Datastore datastore = datastoreOptions.getService();

    EntityQuery entityQueryWithNamespace = Query.newEntityQueryBuilder()
        .setKind("Task")
        .setNamespace("ns1")
        .build();

    EntityQuery entityQueryWithoutNamespace = Query.newEntityQueryBuilder()
        .setKind("Task")
        .build();

    AggregationQuery aggregationQueryWithNamespace = Query.newAggregationQueryBuilder()
        .over(Query.newKeyQueryBuilder().setKind("Task").build())
        .addAggregation(count().as("my_count"))
        .setNamespace("ns1")
        .build();

    AggregationQuery aggregationQueryWithoutNamespace = Query.newAggregationQueryBuilder()
        .over(Query.newKeyQueryBuilder().setKind("Task").build())
        .addAggregation(count().as("my_count"))
        .build();

    System.out.println("Total elements from entity query with namespace " + toList(
        datastore.run(entityQueryWithNamespace)).size());  // 2
    System.out.println("Total elements from entity query without namespace " + toList(
        datastore.run(entityQueryWithoutNamespace)).size()); // 2

    System.out.println("Total elements from aggregation query with namespace " + getOnlyElement(
[...]

PR #1167, opened 2023-08-08
fix: add precondition in RemoteRpc to ensure database ids match

kolea2:


datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java
```diff
@@ -81,7 +89,12 @@ public InputStream call(
       String methodName, MessageLite request, String projectId, String databaseId)
       throws DatastoreException {
     logger.fine("remote datastore call " + methodName);
-
+    Preconditions.checkArgument(
+        databaseIdsMatch(databaseId),
+        "database ids mismatched: request database id: "
+            + databaseId
+            + ", DatastoreOptions db: "
+            + this.databaseId);
```
jainsahab, line 97: Should we make it work in the same way as in #1054 ?

In this case, A databaseId configured on query will be prioritised first, followed by the databaseId on the `DatastoreOptions`.

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/it/ITDatastoreProtoClientTest.java
```diff
@@ -19,32 +19,110 @@
...
+            .build();
+
+    try {
+      DATASTORE.runQuery(request);
+      Assert.fail("should have thrown error");
+    } catch (IllegalArgumentException e) {
+      Truth.assertThat(e.getMessage())
+          .isEqualTo(
+              "database ids mismatched: request database id: new-db, DatastoreOptions db: null");
+    }
```
jainsahab, line 125: nit: can be simplified using `org.junit.Assert#assertThrows(java.lang.Class<T>, org.junit.function.ThrowingRunnable)`.

jainsahab (commented):
a nit and a question.

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/RemoteRpc.java
kolea2: At this point, the request is already built so we can't change the database id (unless we rebuilt it, which would make things a little messier because the request is a `MessageLite` type). Separately, the `google-cloud-datastore` module also behaves this way where it expects the db id to be the same as the DatastoreOptions. Let me know your thoughts.

PR #1178, opened 2023-08-28
fix: checksum format

nimf:
Fix checksum format to match the server's format -- unsigned long.

### DIFF
--- a/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/ChecksumEnforcingInputStream.java
+++ b/datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/ChecksumEnforcingInputStream.java
@@ -76,7 +76,7 @@ class ChecksumEnforcingInputStream extends InputStream {
       endToEndChecksumHandler.update(b, off, i);
     } else {
       // no more payload to read. compute checksum and verify
-      if (!expectedChecksum.equalsIgnoreCase(endToEndChecksumHandler.hash())) {
+      if (!expectedChecksum.equals(endToEndChecksumHandler.hash())) {
         throw new IOException("possible memory corruption on payload detected");
       }
     }
--- a/README.md
+++ b/README.md
@@ -57,13 +57,13 @@ implementation 'com.google.cloud:google-cloud-datastore'
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-datastore:2.16.3'
+implementation 'com.google.cloud:google-cloud-datastore:2.17.1'
 ```
 
 If you are using SBT, add this to your dependencies:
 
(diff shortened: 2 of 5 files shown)


datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/ChecksumEnforcingInputStream.java
```diff
@@ -76,7 +76,7 @@ public int read(byte[] b, int off, int len) throws IOException {
       endToEndChecksumHandler.update(b, off, i);
     } else {
       // no more payload to read. compute checksum and verify
-      if (!expectedChecksum.equalsIgnoreCase(endToEndChecksumHandler.hash())) {
+      if (!expectedChecksum.equals(endToEndChecksumHandler.hash())) {
```
jainsahab, line 79: One QQ,
In this case, client is verifying the response integrity using checksum. IIUC  For older version of this library `expectedChecksum` variable will be in integer (long) format (received as part of response header) but `endToEndChecksumHandler.hash()` will still be returning the checksum in hex representation.

Will it be a breaking change for older version of this library or Am I missing something ?

nimf: It looks like the backend always used or switched to integer format a long time ago. So currently `expectedChecksum` is always in integer format for any client library version. Before this PR `endToEndChecksumHandler.hash()` returns hex format so the feature is currently broken and this PR fixes that.

If we enable the `GOOGLE_CLOUD_DATASTORE_HTTP_ENABLE_E2E_CHECKSUM` with the current client all requests fail because of checksum mismatch (because of different format). But with this PR it works.

jainsahab: Thanks @nimf for the clarification. It makes sense.

I think what we are missing is a separate integration test with `GOOGLE_CLOUD_DATASTORE_HTTP_ENABLE_E2E_CHECKSUM` variable set to `true` so that we can catch these issues sooner, but that can be covered as part of separate PR, I have created #1189  for that. For now it's LGTM.

cc: @kolea2 .

jainsahab (approved):

kolea2:
Looks good, however the lint check is failing. @nimf you can run `mvn com.coveo:fmt-maven-plugin:format` to format the code.

datastore-v1-proto-client/src/main/java/com/google/datastore/v1/client/ChecksumEnforcingInputStream.java
nimf: Thanks! Yes, I'll add it in a separate PR.

PR #1190, opened 2023-09-13
fix: deprecate `databaseId` on datastore-v1-proto-client DatastoreOptions

kolea2:


datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/DatastoreClientTest.java
```diff
@@ -218,13 +218,8 @@ public void create_LocalHost() {
   @Test
   public void setDatabaseId() {
     DatastoreOptions options =
-        new DatastoreOptions.Builder()
-            .projectId(PROJECT_ID)
-            .databaseId("test-db")
-            .localHost("localhost:8080")
-            .build();
+        new DatastoreOptions.Builder().projectId(PROJECT_ID).localHost("localhost:8080").build();
```
jainsahab, line 221: nit: Should we instead keep this test intact until we go to step 2 which is removing the `databaseId` support from `DatastoreOptions` eventually? Coz for the time being we will have deprecated and **untested** code.

jainsahab (approved):

datastore-v1-proto-client/src/test/java/com/google/datastore/v1/client/DatastoreClientTest.java
kolea2: makes sense, thanks! I've reverted this and added a TODO comment to remove it later.

PR #1238, opened 2023-11-16
fix: Fix sync repo setting for V3-experimental (same as main now)

jainsahab:


.github/sync-repo-settings.yaml
```diff
@@ -19,21 +19,24 @@ branchProtectionRules:
       - 'Kokoro - Test: Java GraalVM Native Image'
       - 'Kokoro - Test: Java 17 GraalVM Native Image'
       - javadoc
-  - pattern: 1.106.5-sp
+  - pattern: V3-experimental
```
kolea2, line 22: diff is a little confusing here, can you just edit the required checks for this branch and leave the others as is?

PR #1246, opened 2023-11-29
feat!: Making Datastore implement Closeable

jainsahab:
Noticed the below error in some scenarios
image

With gRPC it has become important to close all the `ManagedChannel`.

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/RetryAndTraceDatastoreRpcDecorator.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/RetryAndTraceDatastoreRpcDecorator.java
@@ -109,6 +109,16 @@ public class RetryAndTraceDatastoreRpcDecorator implements DatastoreRpc {
         () -> datastoreRpc.runAggregationQuery(request), SPAN_NAME_RUN_AGGREGATION_QUERY);
   }
 
+  @Override
+  public void close() throws Exception {
+    datastoreRpc.close();
+  }
+
+  @Override
+  public boolean isClosed() {
+    return datastoreRpc.isClosed();
+  }
+
   public <O> O invokeRpc(Callable<O> block, String startSpan) {
     Span span = traceUtil.startSpan(startSpan);
     try (Scope scope = traceUtil.getTracer().withSpan(span)) {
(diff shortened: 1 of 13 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/spi/v1/DatastoreRpc.java
```diff
@@ -96,4 +96,9 @@ BeginTransactionResponse beginTransaction(BeginTransactionRequest request)
   default RunAggregationQueryResponse runAggregationQuery(RunAggregationQueryRequest request) {
     throw new UnsupportedOperationException("Not implemented.");
   }
+
+  @Override
+  default void close() throws Exception {
```
kolea2, line 101: why make this a default method? Should we just implement in all places?

kolea2 (changes requested):
two general todos:

1. Would be nice to refactor the samples to use a try-with-resources now that this is added. Can you file an issue to track?
2. Maybe add a small test to ensure the client is closed

google-cloud-datastore/src/main/java/com/google/cloud/datastore/spi/v1/DatastoreRpc.java
jainsahab: We can do that, but they will be empty too like this one.

jainsahab: addressed in 4ed8259.

jainsahab:
> two general todos:
> 
> 1. Would be nice to refactor the samples to use a try-with-resources now that this is added. Can you file an issue to track?
> 2. Maybe add a small test to ensure the client is closed

For 1, #1248  is created
For 2, added a test in the latest snapshot.

kolea2 (approved):
lgtm with one question

suztomo:
The "!" in the pull request. Would you add the description why it's breaking change and why it's ok for library users?

suztomo:
I now know the destination branch is "V3-experimental".

google-cloud-datastore/src/main/java/com/google/cloud/datastore/RetryAndTraceDatastoreRpcDecorator.java
```diff
@@ -109,6 +109,16 @@ public RunAggregationQueryResponse runAggregationQuery(RunAggregationQueryReques
         () -> datastoreRpc.runAggregationQuery(request), SPAN_NAME_RUN_AGGREGATION_QUERY);
   }
 
+  @Override
+  public void close() throws Exception {
+    datastoreRpc.close();
```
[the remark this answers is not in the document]
jainsahab: `DatastoreImpl` will catch the exception generated here and print the stack trace.

PR #1275, opened 2024-01-03
feat!: mark interfaces as `@InternalExtensionOnly`

kolea2:
Also replaces current `default` methods as they are no longer required.

More details: https://cloud.google.com/java/docs/reference/api-common/latest/com.google.api.core.InternalExtensionOnly

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
@@ -16,12 +16,14 @@
 
 package com.google.cloud.datastore;
 
+import com.google.api.core.InternalExtensionOnly;
 import com.google.cloud.Service;
 import com.google.datastore.v1.TransactionOptions;
 import java.util.Iterator;
 import java.util.List;
 
 /** An interface for Google Cloud Datastore. */
+@InternalExtensionOnly
 public interface Datastore extends Service<DatastoreOptions>, DatastoreReaderWriter {
 
   /**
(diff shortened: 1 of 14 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
```diff
@@ -505,9 +507,7 @@ interface TransactionCallable<T> {
    * @throws DatastoreException upon failure
    * @return {@link AggregationResults}
    */
-  default AggregationResults runAggregation(AggregationQuery query, ReadOption... options) {
-    throw new UnsupportedOperationException("Not implemented.");
-  }
+  AggregationResults runAggregation(AggregationQuery query, ReadOption... options);
```
cindy-peng, line 510: Hi Kristen, just wondering if the exception is still needed?

kolea2: hi! Since we are removing "default", we cannot add any implementation/exception here. This was used as a workaround previously since we cannot add/update interfaces without adding a 'default' implementation to avoid a breaking change. Since we are now annotating the interface as `@InternalExtensionOnly`, we are able to make such changes and no longer need to have this.

cindy-peng: > hi! Since we are removing "default", we cannot add any implementation/exception here. This was used as a workaround previously since we cannot add/update interfaces without adding a 'default' implementation to avoid a breaking change. Since we are now annotating the interface as `@InternalExtensionOnly`, we are able to make such changes (per the docs, "An interface marked with this annotation can have new methods added to it.") and no longer need to have this.

Got it! That makes sense! Thanks Kristen!

cindy-peng (approved):

bhshkh (approved):

PR #1261, opened 2023-12-13
feat: enabling regapic by giving an option to user to use http

jainsahab:
User can now configure their `Datastore` instance to use http transport.

```java
import static com.google.cloud.datastore.spi.v1.DatastoreRpc.Transport.GRPC;
import static com.google.cloud.datastore.spi.v1.DatastoreRpc.Transport.HTTP;

Datastore datastore = DatastoreOptions.newBuilder().build(); // default gRPC transport
Datastore datastore = DatastoreOptions.newBuilder().setTransport(HTTP).build(); // for http transport
Datastore datastore = DatastoreOptions.newBuilder().setTransport(GRPC).build(); // for gRPC transport
```

PR #1276, opened 2024-01-04
feat: use existing transport options API to set transport

kolea2:
Building on top of https://github.com/googleapis/java-datastore/pull/1261, `setTransportOptions` has always been available as an API to use. Instead of writing custom logic like `setTransport`, we will use the existing API instead and allow it to be set to either gRPC or HTTP.


google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -157,14 +147,18 @@ public DatastoreRpcFactory getDefaultRpcFactory() {
 
     @Override
     public TransportOptions getDefaultTransportOptions() {
-      return getDefaultHttpTransportOptions();
+      return getDefaultGrpcTransportOptions();
```
danieljbruce, line 150: Why does the default transport options need to change?

danieljbruce (approved):
LGTM. Just a question about defaults.

cindy-peng (approved):

PR #1299, opened 2024-01-24
feat: allow setting channel and credential providers in DatastoreOptions

kolea2:
Similar to Firestore, https://github.com/googleapis/java-firestore/blob/main/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java#L162-L168, this allows users to configure more custom options available in gax and gRPC like credentials providers and channel pooling.

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
@@ -103,7 +103,7 @@ public class DatastoreOptions extends ServiceOptions<Datastore, DatastoreOptions
       super(options);
       this.namespace = options.namespace;
       this.databaseId = options.databaseId;
-      this.channelProvider = options.channelProvider;
+      this.channelProvider = validateChannelProvider(options.channelProvider);
       this.credentialsProvider = options.credentialsProvider;
     }
 
@@ -119,11 +119,7 @@ public class DatastoreOptions extends ServiceOptions<Datastore, DatastoreOptions
      *     provider for this client.
      */
     public Builder setChannelProvider(TransportChannelProvider channelProvider) {
-      if (!(channelProvider instanceof InstantiatingGrpcChannelProvider)) {
-        throw new IllegalArgumentException(
-            "Only GRPC channels are allowed for " + API_SHORT_NAME + ".");
-      }
-      this.channelProvider = channelProvider;
+      this.channelProvider = validateChannelProvider(channelProvider);
       return this;
     }
 
@@ -153,6 +149,14 @@ public class DatastoreOptions extends ServiceOptions<Datastore, DatastoreOptions
       this.databaseId = databaseId;
       return this;
     }
+  }
+
+  private static TransportChannelProvider validateChannelProvider(TransportChannelProvider channelProvider) {
+    if (!(channelProvider instanceof InstantiatingGrpcChannelProvider)) {
+      throw new IllegalArgumentException(
+          "Only GRPC channels are allowed for " + API_SHORT_NAME + ".");
+    }
+    return channelProvider;
   }
 
   private DatastoreOptions(Builder builder) {
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
@@ -49,8 +49,8 @@ public class DatastoreOptions extends ServiceOptions<Datastore, DatastoreOptions
   public static final String PROJECT_ID_ENV_VAR = "DATASTORE_PROJECT_ID";
   public static final String LOCAL_HOST_ENV_VAR = "DATASTORE_EMULATOR_HOST";
 
-  private final TransportChannelProvider channelProvider;
-  private final CredentialsProvider credentialsProvider;
+  private final transient TransportChannelProvider channelProvider;
+  private final transient CredentialsProvider credentialsProvider;
 
   private final String namespace;
   private final String databaseId;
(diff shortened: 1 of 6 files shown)


bhshkh (approved):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -84,20 +94,50 @@ public static class Builder extends ServiceOptions.Builder<Datastore, DatastoreO
...
+    private TransportChannelProvider channelProvider = null;
+    private CredentialsProvider credentialsProvider = null;
 
     private Builder() {}
 
     private Builder(DatastoreOptions options) {
       super(options);
       namespace = options.namespace;
       databaseId = options.databaseId;
+      this.channelProvider = options.channelProvider;
```
bhshkh, line 106: Nit: the above 2 variables don't have "this." but the new ones do.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -119,6 +159,32 @@ private DatastoreOptions(Builder builder) {
...
     namespace = MoreObjects.firstNonNull(builder.namespace, defaultNamespace());
     databaseId = MoreObjects.firstNonNull(builder.databaseId, DEFAULT_DATABASE_ID);
+
+    // todo see if we can update this
+    if (getTransportOptions() instanceof HttpTransportOptions
+        && (builder.channelProvider != null || builder.credentialsProvider != null)) {
+      throw new IllegalArgumentException(
+          "Only gRPC transport allows setting of channel provider or credentials provider");
+    } else {
+      this.channelProvider =
```
cindy-peng, line 169: Do we still want to instantiate this to be GrpcTransportProvider if `(getTransportOptions() instanceof HttpTransportOptions) && (builder.channelProvider == null && builder.credentialsProvider == null) == true`?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreOptionsTest.java
```diff
@@ -48,6 +64,16 @@ public void setUp() {
             .setServiceRpcFactory(datastoreRpcFactory)
             .setProjectId(PROJECT_ID)
             .setDatabaseId(DATABASE_ID)
+            .setCredentials(NoCredentials.getInstance())
+            .setHost("http://localhost:" + PORT);
+
+    options2 =
```
cindy-peng, line 70: Is `options2` only for testCustomChannelAndCredentials()? Do we need both `options` and `options2` to cover all the test cases? If not, could we rename them to be more meaningful?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
kolea2: thanks, fixed to have the above two to also have `this.` prefix

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
kolea2: great catch, thank you! I updated the logic and added a test to confirm it's working as expected

google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreOptionsTest.java
kolea2: inlined to the specific test

cindy-peng (approved):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -69,6 +75,10 @@ public ServiceRpc create(DatastoreOptions options) {
         if (options.getTransportOptions() instanceof GrpcTransportOptions) {
           return new GrpcDatastoreRpc(options);
         } else if (options.getTransportOptions() instanceof HttpTransportOptions) {
+          // todo see if we can remove this check
+          if (DatastoreUtils.isEmulator(options)) {
+            throw new IllegalArgumentException("Only GRPC channels are allowed for emulator.");
```
meltsufin, line 80: We use the emulator in Spring Cloud GCP with REST. Maybe [this](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/11c070e2ba0bc17b5ba628ed763bb8d0144ade5a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java#L93) could be helpful?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -84,20 +94,50 @@ public static class Builder extends ServiceOptions.Builder<Datastore, DatastoreO
...
 
     private Builder() {}
 
     private Builder(DatastoreOptions options) {
       super(options);
-      namespace = options.namespace;
-      databaseId = options.databaseId;
+      this.namespace = options.namespace;
+      this.databaseId = options.databaseId;
+      this.channelProvider = options.channelProvider;
```
meltsufin, line 106: Why aren't you using the `setChannelProvider()` here? It has additional checks.

kolea2: `setChannelProvider` returns a Builder, but I can extract out the validation and use it in both places.

kolea2: added!

meltsufin (approved):

PR #1307, opened 2024-01-29
chore: use sdk-platform-java-config to consolidate build configs

mpeddada1:
Notable Changes:
1) Use `gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform*` docker images for Kokoro GraalVM tests  instead of `gcr.io/cloud-devrel-kokoro-resources/graalvm*`.
2) Use `com.google.cloud:sdk-platform-java-config` as the parent which inherits configs from `java-shared-config` and hosts the `google-cloud-shared-dependencies` version under the `google-cloud-shared-dependencies.version` property.  This artifact is versioned to be the same as google-cloud-shared-dependencies.
3) Adjust renovate-bot settings to update docker images when a new version of `sdk-platfrom-java-config` is on Maven Central. 

Example renovate-bot update PR in google-cloud-java: https://github.com/googleapis/google-cloud-java/pull/10290


burkedavison (approved):

pom.xml
```diff
@@ -13,8 +13,8 @@
 
   <parent>
     <groupId>com.google.cloud</groupId>
-    <artifactId>google-cloud-shared-config</artifactId>
-    <version>1.7.1</version>
+    <artifactId>sdk-platform-java-config</artifactId>
+    <version>3.23.0</version>
```
kolea2, line 17: should we also make this a variable/property for this and the above usage?

kolea2 (approved):
lgtm with one nit

pom.xml
mpeddada1: Thanks for the suggestion! I was wondering that too since we have the version specified in two places but unfortunately I don't know if it possible to use a placeholder for the parent since attempting to do so results in this error message: 
```
Properties in parent definition are prohibited
```

I'm guessing Maven tries to determine the parent-child relationships first and then performs property interpolation. Open to any other ideas though.

kolea2: ah okay, no worries - thanks for the explanation!

kolea2:
@mpeddada1 it looks like owlbot reverted the config changes - can you exclude them or should these be added upstream instead?

mpeddada1:
Good catch, @kolea2! Let me try doing a one time update in synthtool: https://github.com/googleapis/synthtool/pull/1921

PR #1360, opened 2024-03-08
test: parameterize tests for http and grpc transports

kolea2:
Also refactors and moves aggregation tests into the main IT

Followup from https://github.com/googleapis/java-datastore/pull/1346


google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/AbstractITDatastoreTest.java
```diff
@@ -1563,6 +1568,285 @@ public void testQueryWithReadTime() throws InterruptedException {
...
+    AggregationQuery aggregationQuery =
+        Query.newAggregationQueryBuilder()
+            .over(baseQuery)
+            .addAggregations(sum("marks").as("total_marks"))
+            .setNamespace(NAMESPACE)
+            .build();
+
+    // sum of 2 entities
+    assertThat(getOnlyElement(datastore.runAggregation(aggregationQuery)).getLong("total_marks"))
+        .isEqualTo(184L);
```
cindy-peng, line 1585: Nit: Since these long int values "184L" and "239L" are used in multiple test cases, can we declare these as constants in the class for reusability?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTestGrpc.java
```diff
@@ -0,0 +1,59 @@
...
+@RunWith(Parameterized.class)
+public class ITDatastoreTestGrpc extends AbstractITDatastoreTest {
+  // setup for default db, grpc transport
+  protected static final RemoteDatastoreHelper HELPER_DEFAULT_GRPC = RemoteDatastoreHelper.create();
+  private static final DatastoreOptions OPTIONS_DEFAULT_GRPC = HELPER_DEFAULT_GRPC.getOptions();
+  private static final Datastore DATASTORE_DEFAULT_GRPC = OPTIONS_DEFAULT_GRPC.getService();
+
+  // setup for custom db, grpc transport
+  private static final RemoteDatastoreHelper HELPER_CUSTOM_DB_GRPC =
+      RemoteDatastoreHelper.create(CUSTOM_DB_ID);
```
cindy-peng, line 35: Will there be any conflict if `CUSTOM_DB_ID` is same for Grpc and Http clients?

cindy-peng (approved):

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTestGrpc.java
kolea2: nope, there shouldn't be! A new namespace is created for each test class run, so they will be running in different ones (within the same db).

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/AbstractITDatastoreTest.java
kolea2: I see your point, but because this is only used in an assertion and not part of the test logic itself, it may be more difficult from a readability perspective to have to check back at the variable name at the top of the file.

PR #1396, opened 2024-04-04
test: add DatastoreException assertions in integration tests

kolea2:


cindy-peng (approved):

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -1620,7 +1620,7 @@ public Integer run(DatastoreReaderWriter transaction) {
       datastore.runInTransaction(callable2, readOnlyOptions);
       fail("Expecting a failure");
     } catch (DatastoreException expected) {
-      assertEquals(3, ((DatastoreException) expected.getCause()).getCode());
```
danieljbruce, line 1623: It looks like `assertDatastoreException` checks for `expected.getCode()`, but this statement checks for `((DatastoreException) expected.getCause()).getCode()`. If these two methods return different things then should we leave this assert statement here? If these two methods are the same thing then why does the expected code change from 3 to 0?

danieljbruce (commented):
Just curious about one thing here.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
kolea2: GREAT catch! I've fixed this, PTAL

danieljbruce (approved):

PR #1244, opened 2023-11-25
feat!: Actionable Error

jainsahab:
Adding the following public methods in `DatastoreException` to provide more info related to errors.
-getReason()
-getDomain()
-getMetaData
-getErrorDetails()

depends on #1240

PR #1409, opened 2024-04-25
feat: have DatastoreException extend BaseHttpServiceException, convert gRPC status codes to existing Datastore codes

kolea2:
Building on top of changes in https://github.com/googleapis/java-datastore/pull/1400, this PR reverts the updates in https://github.com/googleapis/java-datastore/pull/1244 and instead translates gRPC error codes into existing Datastore ones.

This follows the same logic and patterns as in java-storage: https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcToHttpStatusCodeTranslation.java

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
@@ -110,7 +110,7 @@ public final class DatastoreException extends BaseHttpServiceException {
           GrpcToDatastoreCodeTranslation.grpcCodeToDatastoreStatusCode(gsc.getTransportCode());
     }
 
-    // If there is a gRPC exception in our cause change pull it's error message up to be our
+    // If there is a gRPC exception in our cause, pull its error message up to be our
     // message otherwise, create a generic error message with the status code.
     String statusCodeName = statusCode.getCode().name();
     String statusExceptionMessage = getStatusExceptionMessage(apiEx);
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
@@ -16,16 +16,17 @@
 
 package com.google.cloud.datastore;
 
-import static com.google.cloud.BaseServiceException.isRetryable;
-
+import com.google.api.gax.grpc.GrpcStatusCode;
 import com.google.api.gax.rpc.ApiException;
-import com.google.api.gax.rpc.ErrorDetails;
+import com.google.api.gax.rpc.StatusCode;
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryHelper.RetryHelperException;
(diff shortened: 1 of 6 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
```diff
@@ -43,110 +44,106 @@ public final class DatastoreException extends BaseGrpcServiceException {
...
+  static DatastoreException asDatastoreException(ApiException apiEx) {
+    int datastoreStatusCode = 0;
+    StatusCode statusCode = apiEx.getStatusCode();
+    if (statusCode instanceof GrpcStatusCode) {
+      GrpcStatusCode gsc = (GrpcStatusCode) statusCode;
+      datastoreStatusCode =
+          GrpcToDatastoreCodeTranslation.grpcCodeToDatastoreStatusCode(gsc.getTransportCode());
+    }
+
+    // If there is a gRPC exception in our cause change pull it's error message up to be our
```
BenWhitehead, line 113:
```suggestion
    // If there is a gRPC exception in our cause, pull it's error message up to be our
```

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/AbstractITDatastoreTest.java
```diff
@@ -1316,6 +1319,12 @@ public void testUpdate() {
     datastore.delete(ENTITY3.getKey());
   }
 
+  private void assertDatastoreException(
+      DatastoreException expected, String reason, int datastoreStatusCode) {
```
BenWhitehead, line 1323: nit: this naming feels backward to me. I interpret expected as the value that is expected to hold in the assertion, not the exception that was expected to be thrown.

May not be worth the hassell, but I prefer the following pattern for asserting exceptions that are expected to be thrown:
```java
DatastoreException datastoreException = org.junit.Assert.assertThrows(DatastoreException.class, () -> datastore.update(ENTITY3));
com.google.common.truth.Truth.assertThat(datastoreException.getReason()).isEqualTo(NOT_FOUND.name());
com.google.common.truth.Truth.assertThat(datastoreException.getCode()).isEqualTo(5);
```

BenWhitehead (approved):
A couple minor things, that may not need to be addressed in this PR (or at all)

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/AbstractITDatastoreTest.java
kolea2: I like it! I've made these updates.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
kolea2: Fixed in subsequent commit.

danieljbruce (approved):

Issue #1428, opened 2024-05-03
Implement Tracing using OpenTelemetry API in Datastore Java SDK


jimit-j-shah:
This issue tracks the implementation of tracing instrumentation using OpenTelemetry API.

PR #1431, opened 2024-05-03
feat: Adding TraceUtil interface and its implementation to enable Tracing controls via DatastoreOptions

jimit-j-shah:
This change adds:

DatastoreOptions API changes for tracing control using ENABLE_DATASTORE_TRACING flag and DatastoreOptions.
TraceUtil (and its implementations) interface enables interfacing with tracing controls.
OpenTelemetry API and SDK dependencies

Fixes #1428  ☕️

### DIFF
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -16,6 +16,7 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-datastore</site.installationModule>
+    <opentelemetry.version>1.37.0</opentelemetry.version>
   </properties>
   <dependencies>
     <dependency>
@@ -119,12 +120,12 @@
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-api</artifactId>
-      <version>1.37.0</version>
+      <version>${opentelemetry.version}</version>
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-context</artifactId>
-      <version>1.37.0</version>
+      <version>${opentelemetry.version}</version>
     </dependency>
     <!-- END OpenTelemetry -->
 
@@ -180,7 +181,7 @@
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-sdk</artifactId>
-      <version>1.37.0</version>
+      <version>${opentelemetry.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -111,6 +111,14 @@
       <artifactId>jsr305</artifactId>
     </dependency>
 
+    <!-- OpenTelemetry -->
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+      <version>1.37.0</version>
+    </dependency>
+    <!-- END OpenTelemetry -->
+
     <!-- Test dependencies -->
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -117,6 +117,11 @@
       <artifactId>opentelemetry-api</artifactId>
       <version>1.37.0</version>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-context</artifactId>
+      <version>1.37.0</version>
+    </dependency>
     <!-- END OpenTelemetry -->
 
     <!-- Test dependencies -->
@@ -168,46 +173,12 @@
       <version>1.4.2</version>
       <scope>test</scope>
     </dependency>
-    <!-- OpenTelemetry -->
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-sdk</artifactId>
       <version>1.37.0</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-sdk-testing</artifactId>
-      <version>1.37.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-sdk-trace</artifactId>
-      <version>1.37.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-sdk-common</artifactId>
-      <version>1.37.0</version>
-      <scope>test</scope>
-    </dependency>
-    <!-- END OpenTelemetry -->
-    <!-- Cloud Ops -->
-    <dependency>
-      <groupId>com.google.api.grpc</groupId>
-      <artifactId>proto-google-cloud-trace-v1</artifactId>
-      <version>1.3.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.google.cloud.opentelemetry</groupId>
-      <artifactId>exporter-trace</artifactId>
-      <version>0.15.0</version>
-      <scope>test</scope>
-    </dependency>
-    <!-- END OpenTelemetry -->
   </dependencies>
 
   <build>
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -168,6 +168,46 @@
       <version>1.4.2</version>
       <scope>test</scope>
     </dependency>
+    <!-- OpenTelemetry -->
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk</artifactId>
+      <version>1.37.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-testing</artifactId>
+      <version>1.37.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-trace</artifactId>
+      <version>1.37.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-common</artifactId>
+      <version>1.37.0</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- END OpenTelemetry -->
+    <!-- Cloud Ops -->
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-trace-v1</artifactId>
+      <version>1.3.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>exporter-trace</artifactId>
+      <version>0.15.0</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- END OpenTelemetry -->
   </dependencies>
 
   <build>
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -39,6 +39,10 @@
       <artifactId>datastore-v1-proto-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.auth</groupId>
+      <artifactId>google-auth-library-credentials</artifactId>
+    </dependency>
+    <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-api</artifactId>
     </dependency>
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -40,10 +40,6 @@
       <artifactId>datastore-v1-proto-client</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.auth</groupId>
-      <artifactId>google-auth-library-credentials</artifactId>
-    </dependency>
-    <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-api</artifactId>
     </dependency>
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -40,6 +40,10 @@
       <artifactId>datastore-v1-proto-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.auth</groupId>
+        <artifactId>google-auth-library-credentials</artifactId>
+    </dependency>
+    <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-api</artifactId>
     </dependency>
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
@@ -81,11 +81,11 @@ public class DatastoreOpenTelemetryOptions {
     }
 
     /**
-     * Sets the {@link OpenTelemetry} to use with this Firestore instance. If telemetry collection
-     * is enabled, but an `OpenTelemetry` is not provided, the Firestore SDK will attempt to use the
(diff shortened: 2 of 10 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/TraceUtil.java
```diff
@@ -0,0 +1,126 @@
...
+import com.google.api.core.ApiFunction;
+import com.google.api.core.ApiFuture;
+import com.google.cloud.datastore.DatastoreOptions;
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public interface TraceUtil {
+  String ATTRIBUTE_SERVICE_PREFIX = "gcp.datastore.";
```
cindy-peng, line 28: Nit: Add "static final" here since these are constant string values? https://engdoc.corp.google.com/eng/doc/devguide/java/practices/final.md?cl=head#static_fields

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/EnabledTraceUtil.java
```diff
@@ -0,0 +1,301 @@
...
+    @Override
+    public Scope makeCurrent() {
+      try (io.opentelemetry.context.Scope scope = context.makeCurrent()) {
+        return new Scope(scope);
+      }
+    }
+  }
+
+  /** Applies the current Firestore instance settings as attributes to the current Span */
+  private SpanBuilder addSettingsAttributesToCurrentSpan(SpanBuilder spanBuilder) {
```
cindy-peng, line 210: Will these Firestore attributes be documented somewhere?

google-cloud-datastore/pom.xml
```diff
@@ -111,6 +115,19 @@
       <artifactId>jsr305</artifactId>
     </dependency>
 
+    <!-- OpenTelemetry -->
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+      <version>1.37.0</version>
```
kolea2, line 122: nit - can we pull the version from the managed shared dependencies?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
```diff
@@ -0,0 +1,97 @@
...
+public class DatastoreOpenTelemetryOptions {
+  private final boolean enabled;
+  private final @Nullable OpenTelemetry openTelemetry;
+
+  DatastoreOpenTelemetryOptions(Builder builder) {
+    this.enabled = builder.enabled;
+    this.openTelemetry = builder.openTelemetry;
+  }
+
+  public boolean getEnabled() {
```
kolea2, line 32: nit - `isEnabled`

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param enable Whether tracing should be enabled.
+     */
+    @Nonnull
+    public DatastoreOpenTelemetryOptions.Builder setTracingEnabled(boolean enable) {
+      this.enabled = enable;
+      return this;
+    }
+
+    /**
+     * Sets the {@link OpenTelemetry} to use with this Firestore instance. If telemetry collection
```
kolea2, line 84:
```suggestion
     * Sets the {@link OpenTelemetry} to use with this Datastore instance. If telemetry collection
```

kolea2: (and in a few other places)

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -87,6 +107,9 @@ public Builder setTransportOptions(TransportOptions transportOptions) {
 
     @Override
     public DatastoreOptions build() {
+      if (this.openTelemetryOptions == null) {
```
kolea2, line 110: instead of this, you can set `openTelemetryOptions` to the below as the default value, and then allow the setter to override it for a customization.

or, this logic should probably move to the constructor of DatastoreOptions rather than in the `build()` method.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DisabledTraceUtil.java
```diff
@@ -0,0 +1,103 @@
...
+package com.google.cloud.datastore.telemetry;
+
+import com.google.api.core.ApiFunction;
+import com.google.api.core.ApiFuture;
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class DisabledTraceUtil implements TraceUtil {
```
kolea2, line 26: please add some docs explaining what this class does

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/EnabledTraceUtil.java
```diff
@@ -0,0 +1,301 @@
...
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.trace.Tracer;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class EnabledTraceUtil implements TraceUtil {
```
kolea2, line 38: same here, please add some docs

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/TraceUtil.java
```diff
@@ -0,0 +1,126 @@
...
+
+import com.google.api.core.ApiFunction;
+import com.google.api.core.ApiFuture;
+import com.google.cloud.datastore.DatastoreOptions;
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public interface TraceUtil {
```
kolea2, line 27: please mark as `@InternalExtensionOnly`

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/EnabledTraceUtil.java
kolea2: also, is this an internal class? If so, please mark as `@InternalApi`

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -63,17 +69,31 @@ public ServiceRpc create(DatastoreOptions options) {
...
+  public DatastoreOpenTelemetryOptions getOpenTelemetryOptions() {
+    return openTelemetryOptions;
+  }
+
   public static class Builder extends ServiceOptions.Builder<Datastore, DatastoreOptions, Builder> {
 
     private String namespace;
     private String databaseId;
 
+    @Nullable private DatastoreOpenTelemetryOptions openTelemetryOptions = null;
```
kolea2, line 88: would this really be nullable if we're setting it later on?

kolea2 (changes requested):

google-cloud-datastore/src/test/java/com/google/cloud/datastore/telemetry/EnabledTraceUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  EnabledTraceUtil newEnabledTraceUtil() {
+    return new EnabledTraceUtil(getTracingEnabledOptions());
+  }
+
+  @Test
+  public void usesOpenTelemetryFromOptions() {
+    OpenTelemetrySdk myOpenTelemetrySdk = OpenTelemetrySdk.builder().build();
+    DatastoreOptions firestoreOptions =
```
kolea2, line 51:
```suggestion
    DatastoreOptions datastoreOptions =
```

kolea2 (changes requested):

google-cloud-datastore/src/test/java/com/google/cloud/datastore/telemetry/TraceUtilTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Test;
+
+public class TraceUtilTest {
+  @Test
+  public void defaultOptionsUseDisabledTraceUtil() {
+    TraceUtil traceUtil =
+        TraceUtil.getInstance(
+            DatastoreOptions.newBuilder()
+                .setProjectId("test-project")
+                .setDatabaseId("(default)")
```
kolea2, line 31: just for correctness, setting the default database can be done by omitting this line or setting it to "". I would recommend removing this (and below)

kolea2: I would also add: https://github.com/googleapis/java-datastore/blob/grpc-experimental/google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreOptionsTest.java#L58

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -63,17 +69,31 @@ public ServiceRpc create(DatastoreOptions options) {
     }
   }
 
+  @Nonnull
+  com.google.cloud.datastore.telemetry.TraceUtil getTraceUtil() {
```
kolea2, line 73: please add some tests to `DatastoreOptionsTest`

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/EnabledTraceUtil.java
```diff
@@ -0,0 +1,301 @@
...
+
+    public Span(io.opentelemetry.api.trace.Span span, String spanName) {
+      this.span = span;
+      this.spanName = spanName;
+    }
+
+    /** Ends this span. */
+    @Override
+    public void end() {
+      span.end();
```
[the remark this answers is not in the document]
jimit-j-shah: This would leave the span error `UNSET` which means `successful` by default - `OK` is used when we want to explicitly indicate a `success` vs `failure. So this is ok. https://opentelemetry.io/docs/languages/go/instrumentation/#set-span-statuscom/9kz56fmcu2tx6f7

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
jimit-j-shah: Done.

google-cloud-datastore/pom.xml
jimit-j-shah: I did set it up as initially. But hit some dependency issue w.r.t. version. Let me retry.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
jimit-j-shah: Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
jimit-j-shah: Looks like this was a redundant defensive check - removed it as it's already handled in the constructor.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
kolea2: sorry, to clarify - the variable should remain `enabled`, the getter should be renamed to `isEnabled`

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
jimit-j-shah: This could be null i.e. if the user never calls Builder.setOpenTelemetryOptions(). It's the DatastoreOptions.openTelemetryOptions that is @NonNull.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
jimit-j-shah: Done.

cindy-peng (approved):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
```diff
@@ -0,0 +1,97 @@
...
+      return new DatastoreOpenTelemetryOptions(this);
+    }
+
+    /**
+     * Sets whether tracing should be enabled.
+     *
+     * @param enable Whether tracing should be enabled.
+     */
+    @Nonnull
+    public DatastoreOpenTelemetryOptions.Builder setTracingEnabled(boolean enable) {
```
kolea2, line 78: nit - `enabled`

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/EnabledTraceUtil.java
```diff
@@ -0,0 +1,306 @@
...
+            } else if (value instanceof Double) {
+              attributesBuilder.put(key, (double) value);
+            } else if (value instanceof Float) {
+              attributesBuilder.put(key, (float) value);
+            } else if (value instanceof Boolean) {
+              attributesBuilder.put(key, (boolean) value);
+            } else if (value instanceof String) {
+              attributesBuilder.put(key, (String) value);
+            } else {
+              // OpenTelemetry APIs do not support any other type.
```
kolea2, line 157: just a thought - in case this ever changes in the future, would we want to throw a warning rather than an error? and maybe put it in the `attributesBuilder` as an Object type?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
kolea2: Not seeing these, are we missing a commit?

google-cloud-datastore/pom.xml
```diff
@@ -38,6 +39,10 @@
       <groupId>com.google.cloud.datastore</groupId>
       <artifactId>datastore-v1-proto-client</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.auth</groupId>
```
kolea2, line 43: where is this added/used?

kolea2 (approved):
lgtm, have one nit and two questions

google-cloud-datastore/pom.xml
jimit-j-shah: One of my earlier builds had failed with an error for missing `com.google.auth` dependency - so I'd added it then. Maybe my mvn repo wasn't correctly set up. Don't need this anymore.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOpenTelemetryOptions.java
jimit-j-shah: Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/EnabledTraceUtil.java
jimit-j-shah: The intent of this exception is for us to catch this issue before we rollout any changes to public. If we mark it as a warning, I have a feeling this will go unnoticed. And if there are warnings on the client-side, they'll go unheeded too. Besides, we don't expect this list to grow dramatically. WDYT?

google-cloud-datastore/pom.xml
jimit-j-shah: Nvm. The kokoro build did fail. https://github.com/googleapis/java-datastore/actions/runs/9006313814/job/24743556228 I'm going to reinstate the dependency.

PR #1440, opened 2024-05-15
feat: Commit RPC Trace Instrumentation

jimit-j-shah:
- Added end-to-end test for Datastore operations `put`, `add`, `update` and `delete`.
- Updated E2E Test to use the namespace correctly for efficient clean-up of test data

Fixes #1428 ☕️


kolea2 (approved):
I have some comments on the other PR that apply here (refactoring common logic, updating where the namespace is deleted) but will leave them to be addressed there.

PR #1441, opened 2024-05-15
feat: RunQuery trace instrumentation

jimit-j-shah:
- Added test for Datastore.run() method

Fixes #1428  ☕️

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
@@ -242,7 +242,7 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
   com.google.datastore.v1.RunQueryResponse runQuery(
       final com.google.datastore.v1.RunQueryRequest requestPb) {
     com.google.cloud.datastore.telemetry.TraceUtil.Span span =
-        otelTraceUtil.startSpan(com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_RUNQUERY);
+        otelTraceUtil.startSpan(com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_RUN_QUERY);
     ReadOptions readOptions = requestPb.getReadOptions();
     span.setAttribute(
         "isTransactional", readOptions.hasTransaction() || readOptions.hasNewTransaction());
@@ -258,7 +258,7 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
                   : TRANSACTION_OPERATION_EXCEPTION_HANDLER,
               getOptions().getClock());
       span.addEvent(
-          com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_RUNQUERY + ": Completed",
+          com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_RUN_QUERY + ": Completed",
           new ImmutableMap.Builder<String, Object>()
               .put("Received", response.getBatch().getEntityResultsCount())
               .put("More results", response.getBatch().getMoreResults().toString())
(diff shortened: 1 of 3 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -240,20 +241,34 @@ public AggregationResults runAggregation(
...
-          retrySettings,
-          requestPb.getReadOptions().getTransaction().isEmpty()
-              ? EXCEPTION_HANDLER
-              : TRANSACTION_OPERATION_EXCEPTION_HANDLER,
-          getOptions().getClock());
+    com.google.cloud.datastore.telemetry.TraceUtil.Span span =
+        otelTraceUtil.startSpan(com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_RUN_QUERY);
+    ReadOptions readOptions = requestPb.getReadOptions();
+    span.setAttribute(
+        "isTransactional", readOptions.hasTransaction() || readOptions.hasNewTransaction());
```
kolea2, line 248: I recall seeing this in the previous PR as well. Can we extract out this common logic?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITE2ETracingTest.java
```diff
@@ -319,6 +337,7 @@ public void after() throws Exception {
     if (isUsingGlobalOpenTelemetrySDK()) {
       GlobalOpenTelemetry.resetForTest();
     }
+    remoteDatastoreHelper.deleteNamespace();
```
kolea2, line 340: I think this should take place in `@AfterClass` rather than `@After` (unless there is a different namespace per test? But ideally there should be one namespace for the entire suite)

kolea2 (changes requested):

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: That's because both branches were forked from `tracing-main`. I'll rebase them once I'm ready to merge.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITE2ETracingTest.java
jimit-j-shah: Actually this helps reuse of a fixed set of keys in every test case. Are namespaces expensive?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
kolea2: does this PR need rebasing you're saying? Or after?

jimit-j-shah: Oh wait, these changes are specific span attributes per RPC. So you *may* see similar looking span annotations but they're not the same across all RPCs.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITE2ETracingTest.java
jimit-j-shah: Your thoughts on the above comment @kolea2 ?

kolea2: sorry for missing this  - I was thinking we should align with ITDatastoreTest, but if you find it's better to leave this as is for these test cases, that works for me.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
kolea2: Right - I see this in both lookup and runQuery. I think we should extract this duplication out (at least the boolean logic)

PR #1457, opened 2024-05-27
fix: set the correct database id on the key parent when calling Key#getParent

### LINKED ISSUE #1456: Cloud Datastore Java SDK - parent keys are created without the correct database ID
#### Environment details

Java version: 17
SDK version: 2.19.4-SNAPSHOT

#### Steps to reproduce

  1. Create a key with a parent and a non-default database ID
  2. Compare the database ID of the key to the database ID of the parent
  3. Observe the values are different. They should be the same.

#### Code example

```java
Key entityKey = DatastoreOptions.newBuilder()
        .setDatabaseId("my-database-id")
        .build()
        .getService()
        .newKeyFactory()
        .setKind("EntityKind")
        .addAncestor(PathElement.of("ParentKind", 1L))
        .newKey(2L);
System.out.printf("Entity database ID: %s%n", entityKey.getDatabaseId());
System.out.printf("Parent database ID: %s%n", entityKey.getParent().getDatabaseId());
```

nzhenry:
Fixes #1456 ☕️

This is to fix the bug where the database ID is not set correctly on the parent key. The bug causes problems in Objectify by preventing you from fetching an entity from a non-default database, using a parent key.

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/IncompleteKey.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/IncompleteKey.java
@@ -91,9 +91,11 @@ public class IncompleteKey extends BaseKey {
     PathElement parent = ancestors.get(ancestors.size() - 1);
     Key.Builder keyBuilder;
     if (parent.hasName()) {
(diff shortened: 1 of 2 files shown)


kolea2:
Thank you so much for the fix and tests!

kolea2 (approved):
LGTM. Thanks again for finding and fixing this.

PR #1468, opened 2024-05-30
feat: support for transactional operations

jimit-j-shah:
- tested using newTransaction() and runInTransaction()
-  will add more tests in subsequent PR

Fixes #1428  ☕️

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
@@ -105,7 +105,6 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
     private final TransactionCallable<T> callable;
     private volatile TransactionOptions options;
     private volatile Transaction transaction;
-    @Nonnull private final Context parentTraceContext;
 
     ReadWriteTransactionCallable(
         Datastore datastore,
@@ -116,7 +115,6 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
       this.callable = callable;
       this.options = options;
       this.transaction = null;
-      this.parentTraceContext = parentTraceContext;
     }
 
     Datastore getDatastore() {
@@ -139,8 +137,13 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
 
     @Override
     public T call() throws DatastoreException {
-      try (com.google.cloud.datastore.telemetry.TraceUtil.Scope ignored =
-          parentTraceContext.makeCurrent()) {
+      com.google.cloud.datastore.telemetry.TraceUtil traceUtil =
+          datastore.getOptions().getTraceUtil();
+      com.google.cloud.datastore.telemetry.TraceUtil.Span span =
+          traceUtil.startSpan(
+              com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_TRANSACTION_RUN,
+              datastore.getOptions().getTraceUtil().getCurrentContext());
+      try (com.google.cloud.datastore.telemetry.TraceUtil.Scope ignored = span.makeCurrent()) {
         transaction = datastore.newTransaction(options);
         T value = callable.run(transaction);
         transaction.commit();
@@ -167,7 +170,8 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
             com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_TRANSACTION_RUN);
     try (com.google.cloud.datastore.telemetry.TraceUtil.Scope ignored = span.makeCurrent()) {
       return RetryHelper.runWithRetries(
-          new ReadWriteTransactionCallable<T>(this, callable, null, otelTraceUtil.getCurrentContext()),
+          new ReadWriteTransactionCallable<T>(
+              this, callable, null, otelTraceUtil.getCurrentContext()),
           retrySettings,
           TRANSACTION_EXCEPTION_HANDLER,
           getOptions().getClock());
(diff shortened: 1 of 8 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -154,36 +162,41 @@ public T call() throws DatastoreException {
...
+    try (com.google.cloud.datastore.telemetry.TraceUtil.Scope ignored = span.makeCurrent()) {
       return RetryHelper.runWithRetries(
-          new ReadWriteTransactionCallable<T>(this, callable, null),
+          new ReadWriteTransactionCallable<T>(this, callable, null, otelTraceUtil.currentContext()),
           retrySettings,
           TRANSACTION_EXCEPTION_HANDLER,
           getOptions().getClock());
     } catch (RetryHelperException e) {
-      span.setStatus(Status.UNKNOWN.withDescription(e.getMessage()));
+      span.end(e);
```
cindy-peng, line 175: Will there be a problem if we call span.end() again at L178?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -634,10 +647,14 @@ private com.google.datastore.v1.CommitResponse commitMutation(
...
-    com.google.cloud.datastore.telemetry.TraceUtil.Span span =
-        otelTraceUtil.startSpan(com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_COMMIT);
+    final boolean isTransactional =
+        requestPb.hasTransaction() || requestPb.hasSingleUseTransaction();
+    final String spanName =
+        isTransactional
+            ? com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_TRANSACTION_COMMIT
+            : com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_COMMIT;
+    com.google.cloud.datastore.telemetry.TraceUtil.Span span = otelTraceUtil.startSpan(spanName);
     span.setAttribute("isTransactional", requestPb.hasTransaction());
```
cindy-peng, line 657: Does it make sense to update this to `span.setAttribute("isTransactional", isTransactional);`?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -103,13 +105,18 @@ static class ReadWriteTransactionCallable<T> implements Callable<T> {
...
-        Datastore datastore, TransactionCallable<T> callable, TransactionOptions options) {
+        Datastore datastore,
+        TransactionCallable<T> callable,
+        TransactionOptions options,
+        @Nonnull Context parentTraceContext) {
       this.datastore = datastore;
       this.callable = callable;
       this.options = options;
       this.transaction = null;
+      this.transactionTraceContext = parentTraceContext;
```
kolea2, line 119: nit - can we keep the variables and parameters all the same name?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -132,8 +139,9 @@ void setPrevTransactionId(ByteString transactionId) {
 
     @Override
     public T call() throws DatastoreException {
-      transaction = datastore.newTransaction(options);
-      try {
+      try (com.google.cloud.datastore.telemetry.TraceUtil.Scope ignored =
+          transactionTraceContext.makeCurrent()) {
+        transaction = datastore.newTransaction(options);
```
kolea2, line 144: is there a difference in behavior if this is moved to inside the try block rather than before?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -154,36 +162,41 @@ public T call() throws DatastoreException {
...
-    Span span = traceUtil.startSpan(TraceUtil.SPAN_NAME_TRANSACTION);
-    try (Scope scope = traceUtil.getTracer().withSpan(span)) {
+    com.google.cloud.datastore.telemetry.TraceUtil.Span span =
+        otelTraceUtil.startSpan(
+            com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_TRANSACTION_RUN);
+    try (com.google.cloud.datastore.telemetry.TraceUtil.Scope ignored = span.makeCurrent()) {
       return RetryHelper.runWithRetries(
-          new ReadWriteTransactionCallable<T>(this, callable, transactionOptions),
+          new ReadWriteTransactionCallable<T>(
+              this, callable, transactionOptions, otelTraceUtil.currentContext()),
```
kolea2, line 191: sorry for just noticing this now, can we change this to `getCurrentContext()` for consistency?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: This call is idempotent. If `span.end()` has been called before, another `span.end()` will have no effect.
See https://github.com/open-telemetry/opentelemetry-java/blob/73646871e3b3b521d738fdccb7b491ef3f1f628f/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpan.java#L496

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: Good catch. Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: Yes, `newTransaction` sends a `BEGIN_TRANSACTION` RPC to the backend and we want that to get traced.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: Done. Also s/currentSpan/getCurrentSpan

cindy-peng (approved):

PR #1495, opened 2024-06-25
fix: Fixed Span nesting for `ReadWriteTransactionCallable` by using parent SpanContext instead of just parent Context

### LINKED ISSUE #1491: For transaction operations, OpenTelemetry Trace spans not exhibiting correct hierarchy - all operations must be nested under the Run method.
nested span issue for transactions
Thanks for stopping by to let us know something could be better!

jimit-j-shah:
- This fixes the hierarchy of Spans appearing in a transaction under a Run method.
- Tested using existing transaction test

Fixes googleapis/google-cloud-java#11991  ☕️

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
@@ -143,7 +143,7 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
       options = options.toBuilder().setReadWrite(readWrite).build();
     }
 
-    private io.opentelemetry.api.trace.Span startSpanInternal(
+    private io.opentelemetry.api.trace.Span startSpanWithParentContext(
         String spanName,
         com.google.cloud.datastore.telemetry.TraceUtil.SpanContext parentSpanContext) {
       com.google.cloud.datastore.telemetry.TraceUtil otelTraceUtil =
@@ -169,7 +169,7 @@ final class DatastoreImpl extends BaseService<DatastoreOptions> implements Datas
       // debugging. The code below works and is idiomatic but could be prettier and more consistent
       // with the use of TraceUtil-provided framework.
       io.opentelemetry.api.trace.Span span =
-          startSpanInternal(
+          startSpanWithParentContext(
               com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_TRANSACTION_RUN,
               parentSpanContext);
       try (io.opentelemetry.context.Scope ignored = span.makeCurrent()) {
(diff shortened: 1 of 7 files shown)


jimit-j-shah:
Original nested span issue for transactions

Fixed nested span issue for transactions

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -135,26 +143,57 @@ void setPrevTransactionId(ByteString transactionId) {
       options = options.toBuilder().setReadWrite(readWrite).build();
     }
 
+    private io.opentelemetry.api.trace.Span startSpanInternal(
```
cindy-peng, line 146: Nit: The functionality is a bit unclear to me from the name `startSpanInternal` itself. Could we add some comments or rename this function, e.g.`startSpanWithParentContext`?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -165,42 +204,30 @@ public T call() throws DatastoreException {
 
   @Override
   public <T> T runInTransaction(final TransactionCallable<T> callable) {
-    com.google.cloud.datastore.telemetry.TraceUtil.Span span =
```
cindy-peng, line 168: What is the reason that we remove startSpan here and at L221 (edited version) below?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -687,7 +714,7 @@ com.google.datastore.v1.BeginTransactionResponse beginTransaction(
     com.google.cloud.datastore.telemetry.TraceUtil.Span span =
         otelTraceUtil.startSpan(
             com.google.cloud.datastore.telemetry.TraceUtil.SPAN_NAME_BEGIN_TRANSACTION,
-            otelTraceUtil.getCurrentContext());
+            otelTraceUtil.getCurrentSpanContext());
```
cindy-peng, line 717: Do we need to pass the same `otelTraceUtil.getCurrentSpanContext()` to L689?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: Good point. Renamed the function.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: So for now, I've replaced the `TraceUtil.startSpan` with `io.opentelemetry.api.trace.SpanBuilder` - `TraceUtil.startSpan` is really a wrapper around `io.opentelemetry.api.trace.SpanBuilder`.
I need to debug why using `TraceUtil.startSpan` doesn't work as expected (it has probably something to do with how the parent context is stored and propagated via thread-local storage but I'm not sure). 

I don't think this issue is a blocker and I can debug this in the background as we make progress with rest of the tracing work.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
jimit-j-shah: Sorry, I didn't understand the comment.

cindy-peng (approved):

PR #1509, opened 2024-07-02
chore: fix api_description in generation_config

diegomarquezp:


JoeWang1127:
Do you know why there is code change in this PR?

diegomarquezp:
> Do you know why there is code change in this PR?

@JoeWang1127 it's because I only modified api_description while googleapis_committish has been left with a very old value. I'll close this PR and modify the description in the generated PR

PR #1518, opened 2024-07-10
test: Additional Transaction tests and AggregationQuery test

jimit-j-shah:
- pending: span event for AggregationQuery
- added `waitForTracesToComplete` for all tests to reduce/eliminate flakiness caused due to `unflushed` trace spans.
- 
Fixes #1428  ☕️


google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DisabledTraceUtil.java
```diff
@@ -115,6 +116,10 @@ public TraceUtil.Span startSpan(String spanName, TraceUtil.SpanContext parentSpa
     return new Span();
   }
 
+  public SpanBuilder addSettingsAttributesToCurrentSpan(SpanBuilder spanBuilder) {
```
cindy-peng, line 119: Nit: Add `@Override` as EnabledTraceUtil.java

cindy-peng (approved):

PR #1577, opened 2024-09-16
feat: add sample code for multiple inequalities filter

kevkim-codes:
- [x] Make sure to open an issue as a [bug/issue]
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

https://buganizer.corp.google.com/issues/352502736

### DIFF
--- a/samples/snippets/src/main/java/com/example/datastore/filters/MultiIneqFilter.java
+++ b/samples/snippets/src/main/java/com/example/datastore/filters/MultiIneqFilter.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.datastore.filters;
+
+// sample-metadata:
+//   title: Range and inequality filters on multiple properties
+//   description: The following query uses range filters on priority and days to return all tasks with priority greater than four and with less than three days to complete.
+
+import com.google.cloud.datastore.Datastore;
+import com.google.cloud.datastore.DatastoreOptions;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Query;
+import com.google.cloud.datastore.QueryResults;
+import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
+import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
+
+public class MultiIneqFilter {
+  public static void invoke() throws Exception {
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/datastore/filters/MultiIneqFilter.java
```diff
@@ -0,0 +1,59 @@
...
+//   priority and days to return all tasks with priority greater
+//   than four and with less than three days to complete.
+
+import com.google.cloud.datastore.Datastore;
+import com.google.cloud.datastore.DatastoreOptions;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Query;
+import com.google.cloud.datastore.QueryResults;
+import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
+import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
```
minherz, line 31: nit: consider adding another region tag START/END block to include Query, Entity, CompositeFliter, PropertyFilter classes

samples/snippets/src/main/java/com/example/datastore/filters/MultiIneqFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    // Get the results back from Datastore.
+    QueryResults<Entity> results = datastore.run(query);
+
+    if (!results.hasNext()) {
+      throw new Exception("query yielded no results");
+    }
+
+    while (results.hasNext()) {
+      Entity entity = results.next();
+      System.out.printf("Entity: %s%n", entity);
```
minherz, line 56: nit: It is preferred to return the value for "get" example methods. This way, testing the method does not require capturing stdout and avoids risk of flaky testing due to ambiguous content in stdout

samples/snippets/src/test/java/com/example/datastore/filters/MultiIneqFilterSampleIT.java
```diff
@@ -0,0 +1,95 @@
...
+        .set("category", "Personal")
+        .set("done", false)
+        .set("priority", 5)
+        .set("days", 2)
+        .set("description", "Set Up Cloud Datastore")
+        .build();
+
+    datastore.put(task1);
+    datastore.put(task2);
+    datastore.put(task3);
```
minherz, line 77: nit: do we need all 3 entities? minimizing the testing dataset helps to reduce total execution time of the test.

minherz (commented):
Please have a look at a few nit comments.

samples/snippets/src/main/java/com/example/datastore/filters/MultiIneqFilter.java
kevkim-codes: We don't currently have a sample where we use this do we? I'm only adding this region tag as requested by devrel / tech writing.

samples/snippets/src/test/java/com/example/datastore/filters/MultiIneqFilterSampleIT.java
kevkim-codes: This is the data set provided to us, so I think I'd like to keep as is. We're standardizing this across all languages.

samples/snippets/src/main/java/com/example/datastore/filters/MultiIneqFilter.java
kevkim-codes: Done.

PR #1579, opened 2024-09-17
feat: add sample code for multiple inequalities indexing consideration query

cindy-peng:

### DIFF
--- a/samples/snippets/src/main/java/com/example/datastore/filters/IndexingConsiderationQuery.java
+++ b/samples/snippets/src/main/java/com/example/datastore/filters/IndexingConsiderationQuery.java
@@ -28,8 +28,8 @@ import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.QueryResults;
 import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
 import com.google.cloud.datastore.StructuredQuery.Filter;
-import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
 import com.google.cloud.datastore.StructuredQuery.OrderBy;
+import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
 
 public class IndexingConsiderationQuery {
   public static void invoke() throws Exception {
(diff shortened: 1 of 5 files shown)


samples/snippets/src/main/java/com/example/datastore/filters/OrderFieldsQuery.java
```diff
@@ -0,0 +1,62 @@
...
+    Query<Entity> query =
+            Query.newEntityQueryBuilder()
+                    .setKind("employees")
+                    .setFilter(PropertyFilter.gt("salary", 100000))
+                    .setOrderBy(OrderBy.asc("salary"))
+                    .build();
+
+    // Get the results back from Datastore
+    QueryResults<Entity> results = datastore.run(query);
+    // Order results by `experience`
```
daniel-sanche, line 50: should there be some code after this comment?

samples/snippets/src/main/java/com/example/datastore/filters/IndexingConsiderationQuery.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
+
+public class IndexingConsiderationQuery {
+  public static void invoke() throws Exception {
+
+    // Instantiates a client
+    Datastore datastore = DatastoreOptions.getDefaultInstance().getService();
+
+    // Build a query with multi inequal filters and optimized index order of index properties.
+    // [START datastore_query_indexing_considerations]
```
daniel-sanche, line 41: Note: the best practice is for samples to be copy/pasteable, meaning the client and imports are included in the region tag

But IIRC datastore doesn't really follow that convention, so if that's the case, we can just consider that in the future

daniel-sanche (approved):
LGTM, but take a look at my comment on `// Order results by experience`

samples/snippets/src/main/java/com/example/datastore/filters/OrderFieldsQuery.java
cindy-peng: Oh I was assuming users can have the choices of their own to use different approaches for sorting. This sorting logic wasn't provided in the sample on the doc site either: https://cloud.google.com/datastore/docs/multiple-range-fields#java_2
Should we be consistent here?

samples/snippets/src/main/java/com/example/datastore/filters/IndexingConsiderationQuery.java
cindy-peng: Good point. I do see in some of the datastore samples, the client and imports were included in the region tag. I have updated the new region tags to include those.

PR #1595, opened 2024-09-24
fix: Update opentelemetry-sdk dependency to be test-only

jimit-j-shah:
Make opentelemetry-sdk dependency `test`-only and move `RemoteDatastoreHelper` to `test`.


cindy-peng (approved):

jimit-j-shah:
Moved RemoteDatastoreHelper to `test` as it's used only for testing.

PR #1671, opened 2024-11-19
feat: introduce `java.time` methods and variables

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/admin/v1/DatastoreAdminSettings.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/admin/v1/DatastoreAdminSettings.java
@@ -117,7 +117,7 @@ import javax.annotation.Generated;
  *         RetrySettings.newBuilder()
  *             .setInitialRetryDelayDuration(Duration.ofMillis(500))
  *             .setRetryDelayMultiplier(1.5)
- *             .setMaxRetryDelay(Duration.ofMillis(5000))
+ *             .setMaxRetryDelayDuration(Duration.ofMillis(5000))
  *             .setTotalTimeoutDuration(Duration.ofHours(24))
  *             .build());
  * datastoreAdminSettingsBuilder
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/models/ExecutionStats.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/models/ExecutionStats.java
@@ -53,14 +53,14 @@ public class ExecutionStats {
     return debugStats;
   }
 
-  /** This method is obsolete. Use {@link #getExecutionDurationDuration()} instead. */
-  @ObsoleteApi("Use getExecutionDurationDuration() instead")
+  /** This method is obsolete. Use {@link #getExecutionJavaTimeDuration()} instead. */
+  @ObsoleteApi("Use getExecutionJavaTimeDuration() instead")
   public org.threeten.bp.Duration getExecutionDuration() {
-    return toThreetenDuration(getExecutionDurationDuration());
+    return toThreetenDuration(getExecutionJavaTimeDuration());
   }
 
   /** Returns the total time to execute the query in the backend. */
-  public java.time.Duration getExecutionDurationDuration() {
+  public java.time.Duration getExecutionJavaTimeDuration() {
     return executionDuration;
   }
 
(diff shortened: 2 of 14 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/admin/v1/DatastoreAdminSettings.java
```diff
@@ -117,7 +117,7 @@
  *         RetrySettings.newBuilder()
  *             .setInitialRetryDelayDuration(Duration.ofMillis(500))
  *             .setRetryDelayMultiplier(1.5)
- *             .setMaxRetryDelay(Duration.ofMillis(5000))
+ *             .setMaxRetryDelayDuration(Duration.ofMillis(5000))
```
lqiu96, line 120: Hmm, I think this should have been migrated automatically since this is an autogenerated file. Did we miss this in the generator or has datastore not been updated in a while?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/models/ExecutionStats.java
```diff
@@ -51,8 +53,14 @@ public Map<String, Object> getDebugStats() {
...
 
+  /** This method is obsolete. Use {@link #getExecutionDurationDuration()} instead. */
+  @ObsoleteApi("Use getExecutionDurationDuration() instead")
+  public org.threeten.bp.Duration getExecutionDuration() {
+    return toThreetenDuration(getExecutionDurationDuration());
+  }
+
   /** Returns the total time to execute the query in the backend. */
-  public Duration getExecutionDuration() {
+  public java.time.Duration getExecutionDurationDuration() {
```
lqiu96, line 63: Oh I see `DurationDuration` here as well. Hmm, let's see if we can a way to work around this.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/admin/v1/DatastoreAdminSettings.java
diegomarquezp: Hey, yes it may have slipped in as a manual change, but the bot pushed https://github.com/googleapis/java-datastore/pull/1671/commits/8f4e43048878a647a3f13c850e7cbc7433d60f90#diff-7d6e6e7ee7a1678cd72a99d79903fd5bd2164167a5f85ac66159d374f10375fd and "confirmed" that this should use the new API

> it probably hasn't been updated in a while.
True

google-cloud-datastore/src/main/java/com/google/cloud/datastore/models/ExecutionStats.java
diegomarquezp: I have a couple ideas:
 - [ ] `getExecutionDurationJavaTime`
 - [ ] `getExecutionDurationJT`

diegomarquezp: Responded in https://github.com/googleapis/java-storage/pull/2826#discussion_r1854486529

lqiu96: Ok, I think either `getExecutionDurationJavaTime` or `getExecutionJavaTimeDuration` should work. JT might be confusing to know that JT references `java.time`. 

I think `getExecutionDurationJavaTime` actually sounds better since it keeps the original method name together `getExecutionTime`. Let's get Cindy and the Storage SME to agree since I think we should be consistent on the choice.

diegomarquezp: @cindy-peng

cindy-peng (approved):

lqiu96 (approved):
LGTM once we settle `JavaTimeDuration` or `DurationJavaTime`

PR #1698, opened 2025-01-06
feat: Add firestoreInDatastoreMode for datastore emulator

cindy-peng:
**Background**
The newer cloud datastore emulator 2.3.1 supports Firestore in Datastore mode, using the argument --use-firestore-in-datastore-mode: https://cloud.google.com/sdk/gcloud/reference/beta/emulators/datastore/start. We want to introduce the same feature to java-datastore client. More details:
https://cloud.google.com/datastore/docs/tools/datastore-emulator#known_issues

**Changes**
This PR is to introduce "--use-firestore-in-datastore-mode" so datastore client can support datastore emulator with fireStoreInDatastoreMode. The changes is made to LocalDatastoreHelper, which is the class in Java datastore client to manage data emulator life cycle. 

**Challenges**
One issue I ran into is that cloud datastore emulator 2.3.1 which includes `--use-firestore-in-datastore-mode` feature doesn't support Java8. According to Patrick, this was an intentional flip due to Google build environment: https://cloud.google.com/sdk/docs/release-notes#39300_2022-07-12

**Testing changes**
This PR also includes some changes to solve the above issue on Java8 testing VMs.
Currently cloud datastore emulator `Java` env variable to run the emulator package:  https://source.corp.google.com/piper///depot/google3/java/com/google/cloud/datastore/emulator/opensource/cloud_datastore_emulator;l=30. This PR introduced some changes to the testing and building files to export `Java` env vars with Java 11 path from Java8 containers so the tests can run successfully on Java8 environment. This is made possible because Java_11 is also installed on Java8 docker file: https://github.com/googleapis/testing-infra-docker/blob/main/java/java8/Dockerfile#L107

### DIFF
--- a/.cloudbuild/samples_build.yaml
+++ b/.cloudbuild/samples_build.yaml
@@ -14,6 +14,11 @@ steps:
 - name: gcr.io/cloud-devrel-public-resources/java8
   entrypoint: pwd
 - name: gcr.io/cloud-devrel-public-resources/java8
+  entrypoint: echo
+  args: [
+    '${JAVA11_HOME}'
+  ]
+- name: gcr.io/cloud-devrel-public-resources/java8
   entrypoint: bash
   args: [
     '.kokoro/build.sh'
@@ -22,6 +27,7 @@ steps:
   - 'JOB_TYPE=samples'
   - 'GOOGLE_CLOUD_PROJECT=cloud-java-ci-sample'
   - 'KOKORO_GITHUB_PULL_REQUEST_NUMBER=$_PR_NUMBER'
+  - 'SUREFIRE_JVM_OPT=-Djvm=${JAVA11_HOME}/bin/java'
 - name: gcr.io/cloud-devrel-public-resources/java8
   entrypoint: echo
   args: [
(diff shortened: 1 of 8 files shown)


suztomo:
Do you know by any chance why Windows build is affected by the emulator JVM version while the other (ubuntu) Java 8 build is fine?

cindy-peng:
> Do you know by any chance why Windows build is affected by the emulator JVM version while the other (ubuntu) Java 8 build is fine?

The current (ubuntu) Java 8 is already using Java 11 for building however it was not added in Windows. To run cloud datastore emulator 2.31.1, we need either Java 11+ runtime or `Java` environment variable set to be Java 11+: https://source.corp.google.com/piper///depot/google3/java/com/google/cloud/datastore/emulator/opensource/cloud_datastore_emulator;l=30

@suztomo I have made some changes to make the unit tests and integration tests pass on Java8. Could you help take a look and see if that's the right fix?

daniel-sanche (approved):

.cloudbuild/samples_build.yaml
```diff
@@ -22,6 +22,7 @@ steps:
   - 'JOB_TYPE=samples'
   - 'GOOGLE_CLOUD_PROJECT=cloud-java-ci-sample'
   - 'KOKORO_GITHUB_PULL_REQUEST_NUMBER=$_PR_NUMBER'
+  - 'JAVA=/usr/lib/jvm/java-11-openjdk-amd64/bin/java'
```
suztomo, line 25: Can you just use the java11 container for the sample build?

suztomo (approved):

PR #1706, opened 2025-01-09
feat: Improve upon the default gRPC Connection Pool size

cindy-peng:
The current Java SDK uses a default gRPC connection pool with only [ one channel ](https://github.com/googleapis/sdk-platform-java/blob/816b1608e52543b4b5ade40056c89740fe35a459/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L627). This can cause performance issues for clients with high request rates (over 100 queries per second) because they may experience throttling.

To address this, this PR update proposes changing the default connection pool settings to allow for multiple channels (minimum 1, maximum 4). This should improve performance for clients with high request rates, without requiring any changes to their code. Customers can still customize the connection pool settings if needed.

Please refer here for more details: 
[configure a more reasonable default grpc connection pool config](https://b.corp.google.com/issues/378924444)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,21 +59,11 @@ jobs:
         env:
           JOB_TYPE: test
   windows:
-    # Building using Java 8 and run the tests with Java 11 runtime
     runs-on: windows-latest
     steps:
     - name: Support longpaths
       run: git config --system core.longpaths true
     - uses: actions/checkout@v4
-    - uses: actions/setup-java@v4
-      with:
-        java-version: 11
-        distribution: temurin
-    - name: "Set jvm system property environment variable for surefire plugin (unit tests)"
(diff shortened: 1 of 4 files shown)


gkevinzheng (commented):
Could you use constants for gRPC minimum and maximum channel counts?

daniel-sanche (approved):
LGTM, but I agree with Kevin that constants would be good to have

cindy-peng:
> Could you use constants for gRPC minimum and maximum channel counts?

Good catch! Fixed.

gkevinzheng (approved):
LGTM

jimit-j-shah (commented):
Have you tested and verified that this indeed improves latency at the SDK? Running a client with high enough operation qps should be able to verify this. 

Also, you can use tcpdump or netstat to verify the work done over each gcp connection .

PR #1876, opened 2025-06-11
build: disable errorprone for JDK 24+

diegomarquezp:
The current errorprone version is not compatible with JDK 24 onwards. While we dont have a long term solution, we will disable the check for JDK 24.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,9 +233,30 @@
 
   <profiles>
     <profile>
+      <!-- the current errorprone version does not support JDK 24 onwards -->
+      <id>no-errorprone-jdk-24-onwards</id>
+      <activation>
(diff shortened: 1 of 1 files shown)


cindy-peng (approved):

jinseopkim0 (approved):

diegomarquezp:
`error_prone-core` is not managed by sdk-platform-java. It is unique to java-datastore.

We need a long term solution. Two options would be
 - upgrade errorprone core to the latest (which supports JDK 17+ only) and drop the check for Java 8 and 11.
 - Keep this fix as is until the Cloud SDK drops support for JDK 8 and 11

cc: @cindy-peng @blakeli0

PR #1886, opened 2025-06-17
feat: enable grpc configurator for client-side tracing

cindy-peng:


google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
```diff
@@ -222,20 +224,37 @@ private DatastoreOptions(Builder builder) {
...
+            this.traceUtil.getChannelConfigurator();
+        if (channelConfigurator == null) {
+          this.channelProvider =
+              GrpcTransportOptions.setUpChannelProvider(
                   DatastoreSettings.defaultGrpcTransportProviderBuilder()
-                      .setChannelPoolSettings(
-                          ChannelPoolSettings.builder()
-                              .setInitialChannelCount(INIT_CHANNEL_COUNT)
-                              .setMinChannelCount(MIN_CHANNEL_COUNT)
-                              .setMaxChannelCount(MAX_CHANNEL_COUNT)
```
daniel-sanche, line 236: Is there a reason to switch to the grpc default instead of explicitly setting MAX_CHANNEL_COUNT?

daniel-sanche (commented):
Looks good, but I am wondering why the MAX_CHANNEL_COUNT needs to change here

daniel-sanche (approved):

